PR #120, opened 2015-07-16
Remove (minor): Unnecessary and invalid assert check for intervention…

nicolasreed:
… mode of vector model.

It does not seem this assert is needed since the code checks for != in a previous condition. Also, it appears that this assert is missing a '=' for comparing the equality; at the moment, this check seems to want to assign instead of compare values.

### DIFF
--- a/model/Transmission/VectorModel.cpp
+++ b/model/Transmission/VectorModel.cpp
@@ -325,7 +325,6 @@ SimTime VectorModel::initIterate () {
         return sim::zero(); // no initialization to do
     }
     if( initIterations < 0 ){
-        assert( interventionMode = dynamicEIR );
         simulationMode = dynamicEIR;
         return sim::zero();
     }


tph-thuering:
Why do you think the assert comparison would be invalid?

It should be compared with `simulationMode` instead of assigned.
+ the previous condition is a comparison with [`interventionMode`](https://github.com/nicolasreed/openmalaria/blob/minor-assert-removal/model/Transmission/VectorModel.cpp#L323).

I think we rather should change the assignment to a comparison, don't you think?

nicolasreed:
Ah, it makes more sense if `interventionMode` should be `simulationMode`.

So we'd want it to be `assert(simulationMode != dynamicEIR);`?

tph-thuering:
No. I don't think so. I think the assignment should be a comparison. But I haven't reviewed this code yet.

nicolasreed:
Ok, it just doesn't seem to make sense to assert that two values are equal and then assign the variable to the same value, but maybe I'm missing something in the code.

tph-thuering:
Ah, I think I know what you mean.

Better read the code first:

From: https://github.com/SwissTPH/openmalaria/blob/master/model/Transmission/TransmissionModel.h#L42
```
/** Variable describing current simulation mode. */
enum SimulationMode {
    /** Initial mode. Indicates that initialisation still needs to happen (i.e.
     * it is an error if this mode is still set when getEIR is called). */
    preInit,
    /** Equilibrium mode (i.e. just apply a fixed EIR)
     * 
     * This is used for the warm-up period and if we want to separate direct
     * effect of an intervention from indirect effects via transmission
     * intensity. The seasonal pattern and intensity of the EIR do not change
     * over years.
     * 
     * In this mode vector calculations aren't run. */
    forcedEIR,
    /** Transient EIR known
     * 
     * This is used to simulate an intervention that changes EIR, and where we
     * have measurements of the EIR over time during the intervention period.
     */
    transientEIRknown,
    /** EIR changes
     * 
     * The simulation is driven by the EIR which changes dynamically during the
     * intervention phase as a function of the characteristics of the
     * interventions.
     * 
     * Dependending on whether the Vector or NonVector model is in use, this
     * EIR may be calculated from a mosquito emergence rate or be an input EIR
     * scaled by the relative infectiousness of the humans. */
    dynamicEIR,
};
```


From here: https://github.com/SwissTPH/openmalaria/blob/master/model/Transmission/VectorModel.cpp#L328

```
SimTime VectorModel::initIterate () {
    if( interventionMode != dynamicEIR ) {
        // allow forcing equilibrium mode like with non-vector model
        return sim::zero(); // no initialization to do
    }
    if( initIterations < 0 ){
        assert( interventionMode = dynamicEIR );
        simulationMode = dynamicEIR;
        return sim::zero();
    }
```

Do you have a specific bug or why the urge to change this?

nicolasreed:
No bug that I have noticed, but I'd have to write a test to see if this assert statement would ever have caused a problem anyways.

`assert(interventionMode = dynamicEIR);` should either be changed to a comparison (`==` or `!=`) or it should be removed.

I mostly opened this PR because I noticed this line and figured it'd be best to do something about it while I had the file open. :)

tph-thuering:
I would not change anything until I'd understand what it actually does.

I think you are right that an assignment in this assert would result to always true, as the compiler would catch an invalid assignment before the code even could run. But let's better leave it at that, I'll add a TODO comment for review.

PR #127, opened 2015-07-28
Add new --version output

tph-thuering:
This change introduces the new semantic versioning update.

* CMake takes version.txt and generates `model/util/version.h` from `model/util/version.h.in` with the respective semantic version.
* schema-34 will be incompatible with versions before thus set variable `SCHEMA_VERSION_OLDEST_COMPATIBLE` accordingly

@dhardy: What do you think of this?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,7 @@ install:
 
 script:
   - make -j2
+  - ./openMalaria --version
   - test "$TRAVIS_OS_NAME" == "linux" && ctest -j2 || echo "osx test follows"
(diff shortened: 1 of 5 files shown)


dhardy:
@tph-thuering nice idea generating version.h if it reduces the number of locations the version number must be given.

I added `SCHEMA_VERSION_OLDEST_COMPATIBLE` because it used to be the case that new version numbers were introduced for both major and minor changes. But with the new versioning scheme it seems redundant, so do you want to remove it?

tph-thuering:
yes, then I'll remove it.

Thank you for your input, which gives me some confidence :+1:

dhardy:
no problem!

Issue #67, opened 2015-05-08
Meta issue: Automated builds


tph-thuering:
Related issues:  

* #54 
* #66 
* #65

tph-thuering:
For OS X see PR #129: https://github.com/SwissTPH/openmalaria/pull/129

PR #129, opened 2015-07-30
Tidy up .travis.yml + allow multi-os deployment

tph-thuering:


tph-thuering:
OS X builds work now!  #67

PR #193, opened 2017-08-24
enable arm build

mainic00:
Build on raspberry pi (no optimizations, mainly for testing om portal).
Please review before accepting.


dhardy:
You can build the sim like this but probably not match results from other platforms. But maybe that doesn't matter much any more?

mainic00:
Different results from different CPU architectures should not be an issue, even if BOINC is used again at some point in the future:  [Homogeneous redundancy](https://boinc.berkeley.edu/trac/wiki/HomogeneousRedundancy) takes care of this.

PR #198, opened 2017-10-06
Schema 38: allow human-vector interventions to affect fecundity

dhardy:


dhardy:
@tph-thuering what are the above comments doing here?

PR #208, opened 2018-09-13
Update to C++11

dhardy:
Memory management has been updated to use C++11 techniques in some places,
but not in all. A few TODO items have been left.

The update revealed a couple of bugs, likely insignificant (other than
performance).

Results changed in many tests, presumably because some copies were avoided.


dhardy:
I should mention motivation: recently, OpenMalaria has become more difficult to compile on (some) Linux distributions, since the packaged version of Xerces-C is configured for more recent compilers. This should help with that.

Caveat: untested on Windows / VisualC++. I don't know if anyone still relies on this even?

PR #230, opened 2019-03-15
Lognormals

dhardy:
Unifies most log-normal parametrisations, as per email discussions


dhardy:
Interestingly, the test against `example_scenario.xml` fails because it attempts to use the current code (unreleased schema version) while validating the XML against the previous schema 39. This cannot work (if any breaking changes affect the test example, as they do now) yet we also wish the test to use the release schema version when preparing the next schema release — i.e. this test is only relevant when preparing a release. So I guess we'll have to merge this with failed tests for now.

PR #242, opened 2019-07-03
Switch to pcg32 RNG

dhardy:
This switches OpenMalaria to the PCG 64 generator (XSL RR 128 64 variant, aka `pcg64`). This generator has less predictable output than previously-used Mersenne Twister algorithm and performs extremely well in all RNG analysis algorithms I have seen. It is also faster and uses much less memory.

This change also switches to the `boost::random` distribution algorithms which are easier to work with. (We don't use C++11 distribution algorithms since the standard provides no guarantee of reproducibility of results, and the libstdc++ implementation is slower than Boost.) The disadvantage of this is that a Boost update may adjust random algorithm output (see https://stackoverflow.com/questions/56384275/known-differences-in-boost-random-output-from-1-58-to-1-67).


dhardy:
Note: this uses PCG32 instead of PCG64 as originally planned, since there are problems building the latter with MSVC. PCG64 may have better performance.

PR #248, opened 2019-09-02
Hrp2 merge

### LINKED ISSUE #206: hrp2 deletions which make the parasites non-detectable on RDT
There is a lot of interest in hrp2 deletions which make the parasites non-detectable on RDT and their clinical implications and impact on MSAT, FSAT  etc.

The task would be to re-purpose drug-resistance alleles in OM to investigate this. Keep drug sensitivity (IC50) the same for “resistant” and “sensitive” alleles but make the former non-detectable by RDTs. Hopefully the code changes should be relatively minor. 

The function Diagnostic::isPositive in the WithinHost  namespace passes densities but presumably that is total density in the patient i.e. “dens”. Could we have a new drug “locus” like “CQ” or “PPQ” but called “hsp2_deletion” and withinhost tracking  passes total parasite densities ignoring clones with hsp2 deletions when this drug class is present?

dhardy:
Closes #206 

Still needs some documentation in the form of a wiki page


dhardy:
Added a short section in the wiki: https://github.com/SwissTPH/openmalaria/wiki/Diagnostics#hrp2-deletions

@ThomasASmith review?

PR #256, opened 2019-10-11
updateKappa optimisation

dhardy:
Reduces vector model run-times by around 5%, since previously the expensive `WHFalciparum::probTransmissionToMosquito` function was called twice per human per step.

To do this, I had to move the vector model's human population summaries to after the human update step, which affected all the vector-model simulations. I think this change is valid, but wish to investigate further.

Note that this means the effects of TBV vaccines and some interventions affecting vectors take one extra step to kick in. For vaccines I believe this is acceptable (there should be a delay anyway which we don't model, and the NonVector model does this); for vector interventions like ITNs and IRS this may not be ideal. Larviciding interventions should be unaffected (continue to act immediately).

(Alternatively, one can achieve the same thing optimisation by moving all transmission model updates to before human updates, however in this case several outputs like "num transmitting humans", "annAvgK" and EIR outputs are delayed by one step.)


dhardy:
There are some questions to the validity of this change; in any case it is not merely an optimisation. The pure optimisations parts will be moved to a new PR.

PR #258, opened 2019-12-31
RNGs: replace PCG32 with Xoshiro256+

dhardy:
The PCG generator has a significant limitation: different streams are not really independent, thus the 64-bit state + 64-bit stream selector of the PCG32 generator cannot really be considered to have 2^128 independent states. Since we use a randomly-generated state per human simulated, the chance of collision is significant (see Birthday Paradox). This is a more significant problem than would have been the case had PCG64 been used as originally planned (see #242).

Instead, we use Xoshiro256+; see prng.di.unimi.it/
This generator increases the state size to a full 256 bits with no measurable performance cost.

Additionally, this is a 64-bit generator and we take advantage of this to generate the maximum of 53-bit precision `double` values, instead of 32-bit precision (the norm for 32-bit generators).


dhardy:
MSVC is as always generous in the number of compatibility issues resulting. This may or may not be the only one:
```
C:\projects\openmalaria\model\util/xoshiro.hpp(150): error C3688: invalid literal suffix 'p'; literal operator or literal operator template 'operator ""p' not found [C:\projects\openmalaria\build\model\model.vcxproj]
    12>
```
I don't have time to address this now.

acavelan (approved):

contrib/chacha.h
```diff
@@ -122,6 +146,15 @@ inline uint32_t ChaCha<R>::operator()() {
     return block[idx];
 }
 
+template<size_t R>
+inline double ChaCha<R>::gen_double() {
+    // We have a 32-bit integer result. Doubles support 53-bits of precision
+    // (1 bit implied). Use what we have in a single sample (32-bits precision,
+    // lower 21 bits are zero).
+    uint64_t x = this->operator()();
+    return (x << 21) * 0x1.0p-53;
```
acavelan, line 155: Hex float literals are only officially supported in the C++17 standard.

acavelan (commented):
Hexadecimal floating-point literals are only supported in the C++17 standard: https://en.cppreference.com/w/cpp/language/floating_literal

I am able to merge this pull request.
However, using gcc 10.1.0, it will only compile with -std=c++17

I am wondering if we should go for this standard, or if it is still too soon. I am thinking about people wanting to compile OM on their laptop / Ubuntu with a (very?) old versions of GCC.

contrib/chacha.h
dhardy: `1.1102230246251565e-16` resolves to exactly the same value (at least, when printed with 30 decimal places). Shall I just replace the value?

acavelan: The comment above that line states that 'the lower 21 bits are zero'.
I just checked and this is still true for the value you suggest, so I believe it is safe to replace it.
Please go ahead! Thanks!

dhardy: Done

dhardy:
Now the problem is merging... this branch adjusts ALL the expected outputs because the RNG is changed. I guess you did something else also affecting these? Then the only option is to regenerate them.

acavelan:
Thanks a lot! I will handle the merging!

I have changed a few things that affect the simulated EIR as we were trying to improve on that.
I will regenerate the expected results and replace all of them.

PR #272, opened 2020-08-19
Sort anopheles by decreasing annual EIR

acavelan:
sort Anopheles species by decreasing annual EIR to improve the fitting. This leads to a better match between the input EIR and the simulated EIR for some tests.


nakul7:
Approved as discussed. The checks that fail are due to the issues with Boost which will be fixed in version 42.

PR #281, opened 2020-11-03
Improve fitting of the emergence rate

acavelan:
- Greatly improve the fitting of the emergence rate, which causes the simulated EIR to be much closer to the input EIR
- Change the Fourrier coefficients in some tests because the EIR profiles were too steep
- Update the expected results of the tests


melissapenny (approved):
All reviewed with Nakul and Aurelien

PR #289, opened 2020-12-03
force propActive to be either 0 or 1

acavelan:


nakul7 (approved):
As discussed on Tuesday.

PR #293, opened 2021-01-05
Refactor

acavelan:
- remove obsolete EmergenceModels
- uniformisation of TransmissionModel
- merge Simulator into openMalaria.cpp
- switch to C++17, even though Ubuntu Xenial LTS 16.04 is old, it has enough for C++17, which is enough for OpenMalaria.
- remove some global variables
- remove Boost dependency completely
- remove tools, gsl, zlib submodules
- move data/Molineaux to openmalaria-development
- remove easyconfig file
- remove old install scripts
- get rid of std::optional for compatibility with old versions of gcc
- add <algorithm> to Global.h (previously included via boost) to fix sse issues in chacha.h
- add -D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR to CXX flags because C++17 standard got rid of auto_ptr but XSD/Xercec libs still use it (at least on Ubuntu Xenial).
- add cmake AppleClang detection
- make build.sh more POSIX compliant
- change static_assert for compatibility with older versions of gcc


acavelan:
Closing because it has been split into smaller pull requests.

PR #295, opened 2021-01-06
merge Parameters.cpp into Parameters.h

acavelan:
Simplify Parameters definition.


nakul7 (approved):
As discussed on the call.

PR #296, opened 2021-01-06
Refactor main loop

acavelan:
Simplify the main loop and remove global variables.


nakul7 (approved):
As discussed on the call.

PR #298, opened 2021-01-06
Refactor transmission model

acavelan:
Simplify Transmission Model.


nakul7 (approved):
As discussed on call.

PR #299, opened 2021-01-06
Remove boost

acavelan:
Fully remove Boost from OpenMalaria. It is not needed as a dependency anymore.

- switch to C++17, even though Ubuntu Xenial LTS 16.04 is old, it has enough support for C++17
- get rid of std::optional for compatibility with old versions of gcc
- add <algorithm> to Global.h (previously included via boost) to fix SSE compiler errors in chacha.h
- add -D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR to CXX to the flags because C++17 standard got rid of auto_ptr but XSD/Xercec libs still use it
- update static_assert calls for compatibility with older versions of gcc

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,15 +65,9 @@ jobs:
 
     # Max OSX release
     - os: osx
-<<<<<<< HEAD
       osx_image: xcode11.5
       compiler: gcc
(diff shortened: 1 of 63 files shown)


nakul7 (approved):
As discussed.

PR #301, opened 2021-01-11
Add more OS versions to travis tests

acavelan:
- Fixes a bug where in older versions of Mac OS, sometimes Linux, where the output.txt file creation happened after OpenMalaria had returned, causing the tests to not detect the the file in time.


nakul7 (approved):
As discussed

PR #303, opened 2021-01-14
remove extra hidden files

acavelan:
This will remove two files that are not needed in the repo:

- .kdev_include_paths
- .old_travis.yml


nakul7 (approved):
As discussed.

PR #304, opened 2021-01-14
Make non vector model header only

acavelan:
To reduce the number of files overall, to help with the ongoing re-factorization.


acavelan:
Closing because it has been superseeded by the new refactorTransmission pull request.

PR #305, opened 2021-01-15
Update appveyor build

acavelan:
Switch to Visual Studio 2019 for Windows builds because the 2015 is now too old for Cygwin.


nakul7 (approved):
As discussed on the call.

PR #308, opened 2021-03-17
Add errno check

acavelan:
This checks the errno variable every loop iteration instead of just once at the end.


melissapenny (approved):
approved

PR #310, opened 2021-03-17
fix log(0) for false positives while reporting totalDensity

acavelan:
If the diagnostic specificity is less than 1, false positives (humans without infections) will be reported.

When this happens, openMalaria will try to report the log(totalDensity) for the given human. Since this is a false positive, the human has totalDensity = 0, and log(0) sets the errno variable with a "Numerical result out of range".

This change will only report densities if they are higher than 0.


melissapenny (commented):
discussed

PR #311, opened 2021-03-22
add errno check

acavelan:
Check the errno variable every loop instead of once at the end of the simulation.


melissapenny (approved):
thank you!

PR #320, opened 2022-01-07
Standardize VecDay

acavelan:
Part of a series of updates aiming at simplifying the code to ultimately run on GPUs.

- replace the VecDay custom vector object by a standard std::vector


nakul7 (approved):
Approved as discussed

PR #321, opened 2022-01-07
Remove unused code

acavelan:
Remove the following unused files:
- model/util/MultidimSolver.h
- model/util/Uninitialised.h


nakul7 (approved):
Approved as discussed during weekly call.

PR #322, opened 2022-01-07
Streamline non human hosts interventions

acavelan:
Move the nonHumanHosts intervention implementation to the Deployments.
Clean up the Anopheles model implementation.


nakul7 (approved):
Approved as discussed.

PR #323, opened 2022-01-07
Standardize Vector2D

acavelan:
Part of a series of updates aiming at simplifying the code to ultimately run on GPUs.

- replace the Vector2D custom vector object by a standard std::vector
- remove the unused gsl* operators from model/util/vectors.h and model/util/vectors.cpp
- make model/util/vectors.h header only
- remove model/util/vectors.cpp


nakul7 (approved):
Approved as discussed.

PR #324, opened 2022-01-07
Add unit parse

acavelan:
The code in model/misc.cpp is confusing.

It contains both the implementation of the sim::init function (defined in model/sim.h) and unit parsing operators and structures, defined in part in model/timeConversions.h and in part in misc.cpp.

This commit will:
- move unit parsing operators and structures to their own UnitParse.h and UnitParse.cpp files
- move the sim::init function to a new sim.cpp file
- remove the confusing misc.cpp file


nakul7 (approved):
Approved

PR #342, opened 2022-04-01
Fix error when totalDensity is 0 in sumlogDens survey

acavelan:
If the diagnostic specificity is less than 1.0, then there will be false positives, which is expected.

However, OpenMalaria will try to report the sum of the log of the totalDensity values for each host, even when "sumlogDens" is false in the survey optons.

Computing log(totalDensity) when totalDensity is 0 is not valid and raises an errno error, which is then reported in the main().

With this commit, OM will not compute the log if totalDensity is 0.


nakul7 (approved):
Approved as discussed on call.

PR #343, opened 2022-04-01
Fix error when mean is 0 in LogNormalSampler

acavelan:
During the initialization of the LogNormalSampler, mu is set as log(mean), which will raise an error when the mean is 0.

This commit sets mu to -infinity directly when mean is 0, without computing the log first.


nakul7 (approved):
Approved as discussed on call.

PR #353, opened 2022-10-29
add ProbDeathSeeking to VectorPop intervention

acavelan:
Add `probDeathSeeking' parameter effect to the VectorPop intervention. This parameter must be specified as a decay function and allows the user to fix the probability of dying while host seeking.

It can be used as follows:
```
<probDeathSeeking initial="0.01">
        <decay L="3" function="step"/>
</probDeathSeeking>
```

When using this effect, OpenMalaria will dynamically compute the corresponding increase in death rate while host seeking (mu_vA). This is done by internally solving an equation numerically every timestep, and as such a small drop in performance must be expected.

If multiple VectorPop interventions overlap, the cumulative probability will be used. Note that it cannot exceed 1, and OpenMalaria will return an error during the simulation if it ever happens.

### DIFF
--- a/model/Transmission/Anopheles/AnophelesModel.cpp
+++ b/model/Transmission/Anopheles/AnophelesModel.cpp
@@ -38,10 +38,8 @@
 
 struct quadratic_params
 {
-    double ah; // The availability rate of host h. Time-1
-    double an; // The availability rate of host n. Time-1
-    double Nh; // Population size of hosts of type h. Animals.
-    double Nn; // Population size of hosts of type n. Humans.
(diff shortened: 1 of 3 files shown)


nakul7:
This looks good. I suggest replacing the name by probAdditionalDeathSugarFeeding and explicitly stating in the wiki  that this only applies while host-seeking and that it is a daily probability. 

I expect that this would need to be less than 1 since a value of 1 would lead to infinitely high availability rates.

nakul7 (approved):
This is excellent.

PR #354, opened 2022-11-21
Fix het avail not working wih forced eir

acavelan:
* Fix heterogenity of availability not working correctly with the Vector Model
  * The Vector Model now uses the intervention availability instead of the initial availability 
* Add the option of using daily values in the vector model
  * Only works with forced EIR
* Add the option of using the changeEIR intervention with the Vector Model
  * Only works with forced EIR
  * Only works with one species

This update also affects the way EIR is computed in the model. Previously, Fourier coefficients were calculated and stored inside the AnophelesModel where they were used to compute EIR values at the time-step level. Scaling the EIR meant scaling the Fourier coefficients.
With this update, Fourier coefficients are only used during initialization to calculate EIR values at the time-step level and are no longer used in the AnophelesModel. The AnophelesModel now relies on a vector of EIR values at the time-step level and it has become agnostic of what method is used to calculate the EIR values. Scaling the EIR means scaling the vector of EIR values at the time-step level.

Fourier coefficients are only used when:
* Monthly values are used with the smoothing="fourier" option during initialization
* Fitting the simulated EIR to the input EIR after the warmup period with mode="dynamic"
* Initializing EIR values directly from Fourier coefficients
 
This also makes it possible to use daily values directly without smoothing and without Fourier coefficients with the Vector Model. For the same reasons, it is also possible to use the changeEIR intervention, right now limited to one species and forced EIR.


melissapenny (approved):
Approved - rounding errors to unit tests only

PR #356, opened 2022-12-19
add increasing decay functions with complement=true option

acavelan:
Add a "complement" parameter to DecayFunction, which will make the decay function return 1-f(x) instead. Since DecayFunctions assume 0 <= f(x) <= 1.0, this is enough to get an increasing effect where the function will "decay" to 1.0 and instead of 0.0.


melissapenny (approved):
All approved, plots checked

PR #357, opened 2022-12-19
fix Windows builds with Appveyor

acavelan:
cmake is broken in Cygwin: https://github.com/appveyor/ci/issues/3508 and cannot be updated without also upgrading the entire cygwin distribution.

This fix forces a cygwin upgrade and pulls cmake from the official package list instead of updating the cmake package alone.


melissapenny (approved):
Changes approved (fixes windows build fails)

PR #358, opened 2023-01-05
Update GitHub action workflow

acavelan:
- Update the the action checkout to version 3 to avoid a warning
- Remove Ubuntu 18.04 as it is now obsolete


nakul7 (approved):
As discussed

PR #360, opened 2023-01-10
Add new and updated example scenario

acavelan:
This replaces the old `example_scenario.xml` file that came with the release by a new and updated one. In particular the new scenario does not contain any obsolete parameters.

The new scenario also comes with an update plot.py file that is able to put together some simple plots based on the example_scenario.


nakul7 (approved):
As discussed.

PR #364, opened 2023-04-05
add explicit error if seekingDeathrate is 0

acavelan:
The exception is raised if the seeking death rate computed during initialization is negative:

"nopheles::init mosq.seekingDeathRate is <= 0. The probability of a mosquito surviving the entire feeding cycle should be less than the product of the probabilities of a mosquito surviving biting, finding a resting site, resting and ovipositing."


nakul7 (approved):
As discussed.

PR #365, opened 2023-05-03
Performance improvement in DescriptiveInfection::determineDensities()

acavelan:
DescriptiveInfection::determineDensities() is the slowest part of the code base when using the base model.

This rewrites the computation to keep the densities calculations in log space and remove some very slow calls to exp(), pow()  and log().

The performance gain is about 10% on the example scenario and scales with the number of infections. We can expect 20% to 40% in very high transmission settings.

Test outputs are identical.


melissapenny (approved):
All approved, changes and expected time savings

PR #368, opened 2023-05-24
Add advanced decay functions

acavelan:
This is a major refactoring of the decay function implementation. Decay function objects have now entire control on the heterogeneity implementation, which was not the case before. This allows for more complex decay functions to exist.

Decay functions are now stored as unique_ptr. They still need to be sampled via the hetSample method. When doing so, the hetSample method now returns a new (sampled cloned) instance of the decay function, rather than a heterogeneity factor. This allows the decay functions to have their own implementation of heterogeneity, whether it is just sampling and returning a new internal heterogeneity factor or something more complicated depends on the decay function type.

Heterogeneity can be enabled with the CV parameter. Older decay functions sample the half-life of the decay function using a log-normal sampler. Each host then has a different copy of the decay function with a half-life that differs from the mean.

This pull request adds the ability to describe advanced decay function and brings four new decay function types: plus, minus, divides and multiplies, as illustrated below.

Example 1. A biphasic function as the sum of two exponential decays. Note that both the plus, minus, divides and multiplies decay functions expect exactly two other decay functions and they will return an error otherwise. In this example, the other decay functions are added together. This decay function is maxed at 1.0. **Note that we can give an initiallEfficacy to each decay function to weigh their importance. initialEfficay is 1.0 by default and is only really needed for the plus and minus decay functions, but it can be used for all decay functions.**
```
<decay function="plus">
    <decay function="exponential" L="1y" initialEfficacy="0.5"/>
    <decay function="exponential" L="20y" initialEfficacy="0.5"/>
</decay>
```

Example 2. Using s step function to abruptly stop the decay. Here the exponential decay will suddenly stop after 2 years.
```
<effect function="multiplies">
    <decay function="exponential" L="5y" />
    <decay function="step" L="2y" />
</effect>
``` 

Example 3. Recursive definitions are allowed.
```
<decay function="multiplies">
    <decay function="plus">
        <decay function="exponential" L="1y" initialEfficacy="0.5"/>
        <decay function="exponential" L="20y" initialEfficacy="0.5"/>
    </decay>
    <decay function="step" L="25y" initialEfficacy="1.0"/>
</decay>
```

Example 4. Biphasic with heterogeneity.
```
<decay function="plus">
    <decay function="exponential" L="5y" initialEfficacy="1.0" CV="2"/>
    <decay function="exponential" L="10y" initialEfficacy="0.0" CV="2"/>
</decay>
```

In addition, the parameter used to make a decay function increased has been renamed from `complement` to `increasing`.

### DIFF
--- a/model/Transmission/Anopheles/AnophelesModel.cpp
+++ b/model/Transmission/Anopheles/AnophelesModel.cpp
@@ -588,7 +588,7 @@ void AnophelesModel::advancePeriod(double sum_avail, double sigma_df, vector<dou
             continue;
         }
         SimTime age = sim::ts0() - it->deployTime;
-        double decayCoeff = it->availHet.eval(age);
+        double decayCoeff = it->availHet->eval(age);
         leaveRate += it->initialAvail * decayCoeff;
         // sigma_df doesn't change: mosquitoes do not survive traps
         it++;
--- a/model/Transmission/Anopheles/AnophelesModel.cpp
+++ b/model/Transmission/Anopheles/AnophelesModel.cpp
@@ -588,7 +588,7 @@ void AnophelesModel::advancePeriod(double sum_avail, double sigma_df, vector<dou
             continue;
         }
         SimTime age = sim::ts0() - it->deployTime;
-        double decayCoeff = trapParams[it->instance].availDecay->eval(age, it->availHet);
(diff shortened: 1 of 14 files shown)


nakul7 (approved):
approved.

PR #369, opened 2023-06-15
add Gamma distribution for entoAvailability

acavelan:
Add the possibility to sample the host availability from on the following distributions:
- const (default)
- lognormal with CV or variance
- gamma with CV or variance

It can be specified via the <availability ... /> parameter in the entomology section of the xml, for each species.

Some examples include:
```
<availability /> <!-- default -->
<availability distr="gamma" CV="2"/>
<availability distr="lognormal" CV="2"/>
<availability distr="gamma" variance="3"/>
<availability distr="lognormal" CV="3"/>```


nakul7 (approved):
As discussed.

PR #371, opened 2023-07-18
Refactoring

acavelan:
The goal of this pull request is to simplify the code, to make it more explicit in places and to remove hidden layers that add complexity. It also relies on new features of C++.

This targets mostly the Population, Human, HumanHet and PerHost classes.

The FastDelegate dependency has been replaced by the standard std::function.

Other parts of the code have been simplified:
- DocumentLoader
- Transmission initDuration, preInitDuration
- main


nakul7 (approved):
As discussed

PR #372, opened 2023-07-18
remove old doc folder that has been superseded by the wiki

acavelan:
This removes the 'doc' folder from the repository. As far as I can see, the folder contains old documentation. The documentation should be in the wiki. This cleans up the top folder.


nakul7 (approved):
As discussed in call.

PR #376, opened 2023-09-11
remove text output in LinearDecayFunction

acavelan:
An unnecessary text output was left in the LinearDecayFunction, this pull request removes it.


nakul7 (approved):
approvied

PR #377, opened 2023-12-14
Add --progress command line option and remove progress updates by default

acavelan:
This will disable progress outputs, e.g. 85%, 86%, ... **by default**, but not error outputs or warnings. 

Add the --progress command line option to get the progress output back.

The reason for this is performance. When users are running large simulations with many OpenMalaria instances on the same computing node, all the outputs are creating a congestion on the file system.

Eventually, the standard output (which contains the progress outputs) is written to a file and when hundreds of instances are trying to write to a file every iteration this creates a congestion. The shared file system on any computing node is slow and does not work in parallel so each instance has to wait for the file system to be available.

We observed a loss of efficiency of up to 50%, so with this pull request we can expect up to a 50% improvement in performance on large simulations.

### DIFF
--- a/model/main.cpp
+++ b/model/main.cpp
@@ -119,7 +119,8 @@ void run(Population &population, TransmissionModel &transmission, SimTime humanW
 
         sim::end_update();
 
-        print_progress(lastPercent, estEndTime);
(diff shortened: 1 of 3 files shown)


nakul7 (approved):
As discussed on call.

PR #379, opened 2024-01-03
Add exception when emergence is negative in SimpleMPD model

acavelan:
When the seasonality increases too fast and when the annual EIR is low, the following quantity can become negative:

`            invLarvalResources[t] = (probPreadultSurvival * yt - mosqEmergeRate[t]) / (mosqEmergeRate[t] * yt);
`
which causes the emergence rate to be negative and crash the simulation.

This pull request does not fix the error, but the code will now throw a helpful exception and stop the simulation when this happens.  This error is rare and can usually be fixed by adjusting the seasonal monthly values or the fourier coefficients and by making sure that the lowest EIR is within 1% of the peak EIR.

### DIFF
--- a/model/Transmission/Anopheles/SimpleMPDAnophelesModel.h
+++ b/model/Transmission/Anopheles/SimpleMPDAnophelesModel.h
@@ -103,8 +103,8 @@ public:
     {
         bool fitted = AnophelesModel::initIterate();
 
-        int y1 = sim::oneYear(), y2 = sim::fromYearsI(2), y3 = sim::fromYearsI(3), y4 = sim::fromYearsI(4),
-                y5 = sim::fromYearsI(5);
+        int y1 = sim::oneYear(), y2 = sim::fromYearsI(2), y3 = sim::fromYearsI(3), y4 = sim::fromYearsI(4), y5 = sim::fromYearsI(5);
+
         assert((int)mosqEmergeRate.size() == y1);
 
         for (int t = 0; t < y1; t++)
@@ -123,6 +123,11 @@ public:
 
     virtual double getEmergenceRate(const SimTime &d0, const std::vector<double> &mosqEmergeRate, double nOvipositing)
     {
+        // // Get emergence at start of step:
+        // SimTime dYear1 = mod_nn(d0, sim::oneYear());
+        // // Simple model: fixed emergence scaled by larviciding
+        // return mosqEmergeRate[dYear1];
+
         // Simple Mosquito Population Dynamics model: emergence depends on the
         // adult population, resources available, and larviciding.
         // See: A Simple Periodically-Forced Difference Equation Model for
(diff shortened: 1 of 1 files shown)


ThomasASmith:
why not replace the emergence rate with zero and send a warning, without stopping the simulation

acavelan:
@ThomasASmith This is exactly what I did initially. However setting the emergence to 0 will affect the simulation and because it is quite rare, users running thousands of scenarios will not see it unless they manually check all the logs.

By throwing an exception, the user will see that an output is missing and can decide how to handle the situation. 

In most cases I would argue this error is caused by a bad or unrealistic seasonality. In addition, the Fourier transform that we use to smooth the input EIR values struggles with small values. It seems that just setting the lowest EIR values to 1% of the peak will solve all the problems.

I actually have another open pull request that will automatically set the minimum EIR to 1% of the max EIR in all scenarios but this is not always needed so it has been pending for a while. Maybe it's time to merge this one too?

Many users probably have this problem without realizing their smoothed input EIR in OpenMalaria doesn't actually match their initial monthly values. 

To illustrate this, you can copy-paste the following monthly EIR values into this tool: https://swisstph.github.io/openmalaria/fourier The following values will crash the simulation when using the SimpleMPD model. You can see with the tool that the Fourier series does not match the input very well.

```
1.0
0.950484
0.850176
0.998646
0.592984
0.090076
0.002824
0.001974
0.019501
0.115986
0.440608
0.866085
```
image

However, setting the two smallest values to 0.01 or moving the decimal place solves the problem:
```
1.0
0.950484
0.850176
0.998646
0.592984
0.090076
0.02824
0.01974
0.019501
0.115986
0.440608
0.866085
```
image

Not only the smoothed EIR matches the monthly values, but the simulation no longer crashes with the SimpleMPD model.

ThomasASmith:
Thanks for the explanation. You have obviously thought it through.
Tom

From: Aurélien Cavelan ***@***.***>
Sent: Wednesday, January 3, 2024 8:24 PM
To: SwissTPH/openmalaria ***@***.***>
Cc: Thomas A. Smith ***@***.***>; Mention ***@***.***>
Subject: [EXTERN] Re: [SwissTPH/openmalaria] Add exception when emergence is negative in SimpleMPD model (PR #379)


@ThomasASmith<https://github.com/ThomasASmith> This is exactly what I did initially. However setting the emergence to 0 will affect the simulation and because it is quite rare, users running thousands of scenarios will not see it unless they manually check all the logs.

By throwing an exception, the user will see that an output is missing and can decide how to handle the situation.

In most cases I would argue this error is caused by a bad or unrealistic seasonality. In addition, the Fourier transform that we use to smooth the input EIR values struggles with small values. It seems that just setting the lowest EIR values to 1% of the peak will solve all the problems.

I actually have another open pull request that will automatically set the minimum EIR to 1% of the max EIR in all scenarios but this is not always needed so it has been pending for a while. Maybe it's time to merge this one too?

Many users probably have this problem without realizing their smoothed input EIR in OpenMalaria doesn't actually match their initial monthly values.

To illustrate this, you can copy-paste the following monthly EIR values into this tool: https://swisstph.github.io/openmalaria/fourier The following values will crash the simulation when using the SimpleMPD model. You can see with the tool that the Fourier series does not match the input very well.

1.0

0.950484

0.850176

0.998646

0.592984

0.090076

0.002824

0.001974

0.019501

0.115986

0.440608

0.866085

image.png (view on web)<https://github.com/SwissTPH/openmalaria/assets/2157755/155954ef-af33-4933-a9c1-19e677f6b8b5>

However, setting the two smallest values to 0.01 or moving the decimal place solves the problem:

1.0

0.950484

0.850176

0.998646

0.592984

0.090076

0.02824

0.01974

0.019501

0.115986

0.440608

0.866085

image.png (view on web)<https://github.com/SwissTPH/openmalaria/assets/2157755/714d9e1c-8549-4897-9263-2a443cc9f16c>

Not only the smoothed EIR matches the monthly values, but the simulation no longer crashes with the SimpleMPD model.

nakul7 (approved):
As discussed on call

PR #388, opened 2024-06-18
update hsMemory to report new epidodes when time passed = hsMemory instead hsMemory + 1t

acavelan:
The clinical element currently simply has a healthSystemMemory attribute: if, when a sickness bout occurs, the previous episode was less than or equal to this many time-steps ago, the new bout is considered a second case (second-line treatment may be used and the two bouts will likely be reported as a single episode). The healthSystemMemory is defined in time steps.

Currently in the 5 day model, a value of 6 corresponds to 35 days (not the 30 days previously claimed).

This pull request will make a value of 6 correspond to 30 days.


melissapenny (approved):
Discussed and approved

PR #389, opened 2024-06-18
Update example scenario

acavelan:
By default, the DecisionTree health system redirects both first line and second line case types to the UCOfficial decision tree. Adding a CaseType node is necessary to handle them separately.

Changing latentp to "15d" removes a warning and is the same as "3" (time-steps).


melissapenny (approved):
Required to differentiate between first and second line treatment

PR #392, opened 2024-07-12
add model option for health system memory fix, true by default

acavelan:
Add the following model option:

```
HEALTH_SYSTEM_MEMORY_FIX
```

This option is set to true by default. It must be set to false if used with the current base model:

```
<option name="HEALTH_SYSTEM_MEMORY_FIX" value="false"/>
```

This makes the health system memory behaves as expected. With this option, specifying a health system memory of 30 days means that a new clinical case will be reported as early as 30 days after the initial case (6 time-steps), rather than 35d (7 time-steps) currently. With this option, setting the health system memory to 5 days allows the model to report a new clinical case every time step.

### DIFF
--- a/model/Clinical/Episode.cpp
+++ b/model/Clinical/Episode.cpp
@@ -40,7 +40,14 @@ void Episode::flush() {
 
 void Episode::update (const Host::Human& human, Episode::State newState)
 {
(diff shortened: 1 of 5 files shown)


nakul7 (approved):
As discussed

PR #393, opened 2024-07-12
Schema-47.0

acavelan:
Update version to 47.0.

New Elimination model, see: #384

The following model option is now needed to run the base model:
<option name="HEALTH_SYSTEM_MEMORY_FIX" value="false"/>

Github CI workflow updated to run on:
- MacOS 12, 13, 14
- Ubuntu 22.04, 24.04

Example scenario updated with DecisionTree5day firstLine/secondLine example, correct model options and latentp parameter.

DecisonTree5Day updated:
- New `<severe>` 
- New `<cohort component="id">`
- `<fever loopBack="5d">` changed to `<uncomplicated memory="5d">` (Note that the memory must be smaller than the healthSystemMemory parameter, but is now independent from it otherwise).

Change CMake minimum version to 3.5 to avoid warnings during compilation.

### DIFF
--- a/model/util/DocumentLoader.h
+++ b/model/util/DocumentLoader.h
@@ -34,7 +34,7 @@ namespace OM
 { 
     namespace util
     {
-        static const int SCHEMA_VERSION = 46;
+        static const int SCHEMA_VERSION = 47;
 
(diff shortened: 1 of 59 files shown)


nakul7 (approved):
Approved as discussed

PR #399, opened 2024-07-25
update ffp-model flag for new intel compiler

acavelan:
The new intel compiler no longer supports "source" for the ffp-model. 
Using "strict" instead to enforce IEEE floating-point compliance.


melissapenny (approved):
thank you!

PR #400, opened 2024-12-10
update vivax clinical model

rossaman4:
Changes to the files WHvivax.cpp, WHvivax.h and vivax.xsd
- some bugs removed (previous version of WHvivax.cpp was broken)
- probability of clinical illness updated 
- different relapse classifications introduced (via vivaxClinOption, variants are related to whether the relapse can trigger illness or not based on whether there is a blood-stage infection by the same brood at the same time and whether the primary infection caused illness )
- Corresponds to manuscript "The probability of acute P vivax illness following primary infection and relapse in Papua New Guinea" (vivaxClinOptions: A1j, A2j, B1j, B2j)
- nb This is not currently a full vivax model


acavelan (approved):
Reviewed together with @rossaman4

acavelan:
All checks OK. macos-12 is deprecated so the test was cancelled.

PR #403, opened 2025-01-16
fix memory leak from missing destructor in Sampler

acavelan:
This fixes a small memory leak that happened at the end of the program.
This leak had no impact on the simulation or on available memory.


nakul7 (approved):
Approved

Issue #233, opened 2019-05-23
Issue with drug conversion model


sharma-raman:
I have hit an issue when evoking the drug conversion model for Artemether-DHA.

Please find the link to a Gist for the corresponding XML input file:

https://gist.github.com/sharma-raman/cd87a7ebfc445b189d8be2e645ca9583

I am getting the following error message:
```
BoincWrapper: not using BOINC
^M[  0%]        Code error: calculateFactor: error epsilon is large: 0.102515 (integral is 13.4196)
Call stack, starting from /home/sharmar/Documents/src/openmalaria-master_v40_debug/model/PkPd/Drug/LSTMDrugConversion.cpp:156:
+0x39e  OM::PkPd::LSTMDrugConversion::calculateDrugFactor(OM::WithinHost::CommonInfection*, double) const
+0x43   OM::PkPd::LSTMModel::getDrugFactor(OM::WithinHost::CommonInfection*, double) const
+0x25a  OM::WithinHost::CommonWithinHost::update(int, std::vector<double, std::allocator<double> >&, double, double)
+0x144  OM::Host::Human::update(bool)
+0xb8   OM::Population::update1(OM::SimTime)
+0x29d  OM::Simulator::start(scnXml::Monitoring const&)
+0x188  main()
+0xe7   __libc_start_main()
+0x2a   _start()
This is likely an error in the C++ code. Please report!
OpenMalaria: No such file or directory
```
I defined two AR drugs (AR_50fold and AR_sens) with conversion to two separate DHA metabolite definitions but only used AR_sens in the health system. This was to test whether we can use the  drug conversion model with dropping in resistances at some point. 

I also tried to a simplified version with just AR_sens defined (remove all definitions associated with AR_50fold), to check if the multiple definitions of drug conversions confused things and I got the same type of error:
```
rm: cannot remove 'timer.txt': No such file or directory
rm: cannot remove 'output.txt': No such file or directory
BoincWrapper: not using BOINC
^M[  0%]        Code error: calculateFactor: error epsilon is large: 0.102515 (integral is 13.4196)
Call stack, starting from /home/sharmar/Documents/src/openmalaria-master_v40_debug/model/PkPd/Drug/LSTMDrugConversion.cpp:156:
+0x39e  OM::PkPd::LSTMDrugConversion::calculateDrugFactor(OM::WithinHost::CommonInfection*, double) const
[...]

PR #404, opened 2025-02-05
decrease abs_eps and rel_eps in gsl_integration_qag

acavelan:
The estimated error from the integral calculation in the PkPD model is sometimes too high, and the code returns an error. 

This pull request decreases the error threshold, in other words, it forces the calculation to be more precise.


b-s-code (commented):
Decreasing the error limits we pass to the gsl call and then updating the tests seems reasonable to me.

Two suggestions:

1. While we're already changing code here, adding some whitespace before this function would improve readability.  The comment about memory leakage does not appear to apply to `LSTMDrugThreeComp::calculateFactor` (which, as I understand it, gets called during time step updates - not just "at end of program").  But the comment's proximity to this function caused me some initial confusion.

1. For my own understanding, can you please add a little more detail on the problem this PR fixes?  Specifically, does "the code returns an error" mean that openMalaria crashes, a message gets written to stderr then execution continues, or something else?

acavelan:
Very reasonable comment. I added a white space before the function.

This function is part of the PkPd model in OpenMalaria (see: https://github.com/SwissTPH/openmalaria/wiki/ScenarioPharmacology).

One user recently experienced the following error:
```Code error: calculateFactor: error epsilon is large: 0.055286 (integral is 5.60774)
Call stack, starting from /scratch/soft/easybuild/OpenMalaria/47.0/intel-compilers-2023.1.0/openmalaria-schema-47.0/model/PkPd/Drug/LSTMDrugThreeComp.cpp:151:
	()
	()
	()
	()
	()
	()
	()
+0x29d90	()
+0x80	__libc_start_main()
	()
This is likely an error in the C++ code. Please report!
```

Now this is the part of the code I'm the least familiar with, but it looks like this is an error in the numerical integration of the drug killing factor (related: https://github.com/SwissTPH/openmalaria/issues/233).

Testing with lower error limits forced the calculation to be more precise and within the original bounds at no noticeable cost.

b-s-code (approved):
Thanks for clarifying.

Changes look good to me 👍

PR #402, opened 2025-01-09
add <infectionOrigin> to the DecisionTree

acavelan:
Add `<infectionOrigin>` together with `<imported>`, `<introduced>` and `<indigenous>` branches to the DecisionTree.

Users can apply different treatments or deploy different interventions depending on if the host classifies as  "imported", "introduced" or "indigenous" based on the origin of its current infection(s). 

If the DecisionTree is used in an intervention, or if NonMalariaFevers are enabled in the healthSystem, then the user should also do a diagnostic to ensure that the host is infected before using the <infectionOrigin> node. If the host is not infected, the node will always execute the <indigenous> branch as this is the default infection origin.

Note that imported and introduced cases only exist when imported infections are enabled. This is done via the `<importedInfections>` intervention, for example:

```xml
<importedInfections name="5 infections per individuals per year">
<timed>
  <rate time="0" value="5000"/> <!-- infections per 1000 individuals per year -->
</timed>
</importedInfections>
```

Example of healthSystem:
```xml
<!-- The more flexible DecisionTree5Day replaces the older ImmediateOutcomes. 
The probabilities below are greatly affected by the presence (or the absence) of a clinic or hospital on site -->
<DecisionTree5Day name="example name">
    <pSeekOfficialCareUncomplicated1 value="0.04"/><!-- 5-day probability that a patient uncomplicated case seeks official care -->
    <pSelfTreatUncomplicated value="0.0"/><!-- 5-day probability that a patient with uncomplicated case will self-treat -->
    <pSeekOfficialCareUncomplicated2 value="0.04"/><!-- 5-day probability that a patient with recurrence seeks official care -->
    <pSeekOfficialCareSevere value="0.48"/><!-- 5-day probability that a patient with severe case seeks official care -->
    <treeUCOfficial>
        <caseType>
            <firstLine>
                <infectionOrigin>
                    <imported>
                         <treatSimple durationLiver="0" durationBlood="1t"/>
                    </imported>
                    <introduced>
                         <treatSimple durationLiver="0" durationBlood="1t"/>
                    </introduced>
                    <indigenous>
                         <treatSimple durationLiver="0" durationBlood="1t"/>
                    </indigenous>
                </infectionOrigin>
            </firstLine>
            <secondLine>
                <infectionOrigin>
                    <imported>
                         <treatSimple durationLiver="0" durationBlood="1t"/>
                    </imported>
                    <introduced>
                         <treatSimple durationLiver="0" durationBlood="1t"/>
                    </introduced>
[...]

PR #407, opened 2025-03-18
Fix infection origin: remove unnecessary UC check

acavelan:
Fix related to: https://github.com/SwissTPH/openmalaria/pull/402

The UC check is not necessary for InfectionOrigin. This can be used both inside an intervention and for Uncomplicated Care in the health system.


b-s-code (approved):
(Happy for this PR to be merged regardless of what the answers to the above questions may be.)

acavelan:
Exactly, the DecisionTree is defined once in the xml (in a recursive manner) and the same definition is used in both the intervention section and the health system section. So it's easier to work with this one definition and enforce the extra rule in the C++ code.

The CaseType is defined as: "A switch which chooses a branch deterministically, based on whether the patient was treated recently (second line) or not (first line). For uncomplicated cases only." This only works for health system treatments, not treatments given through interventions, which are a bit more flexible.

b-s-code:
Makes sense - thanks for clarifying.

PR #408, opened 2025-03-18
Fix: Update minimum CMake version requirement to 3.15

acavelan:
Fix a warning with CMake.


nakul7 (approved):
As discussed for the small update.

PR #412, opened 2025-05-13
cleanup: Remove include attribute for <parameter> elements in XML schema.

b-s-code:
# Problem

Some minor cleanup is needed where model parameters are defined in OpenMalaria's input XML file.

The `include` attribute on `<parameter>` elements is not used by the simulation at all.

The attribute is a relic from when OpenMalaria used to run on a volunteer computing platform, which has not been a supported platform since https://github.com/SwissTPH/openmalaria/pull/236 .

# Solution

1. Removed the `include` attribute from the example scenario XML file.
2. Removed the `include` attribute from the XML schema.
3. Removed the `include` attribute from nearly all the XML files used by tests.
   -  To do this I ran `sed -i -E 's/include=\"(0|1|false|true)\" //' *.xml` then manually inspected all the results by eye.

I did not edit `unittest/scenario.xml`, which has not been touched in 16 years and appears to be using an old schema version anyway.

It is worth noting this PR will make it so that users have to remove all attribute specifications like the following from all of the `<parameter>` elements in any existing XML files they wish to use with newer OpenMalaria versions.

`include="0"`,
`include="1"`,
`include="false"`
`include="true"`

For example, `<parameter include="0" name="'-ln(1-Sinf)'" number="1" value="0.050736"/>` would have to be changed to `<parameter name="'-ln(1-Sinf)'" number="1" value="0.050736"/>`.

The error message a user will see if any `include` attributes remain present in in their input XML file is shown below.  The error message is imperfect since a user may interpret "not declared" as meaning that there is some necessary thing missing from their `parameter` element.  But I think the message is acceptable.

```
:341:101 error: attribute 'include' is not declared for element 'parameter'
```

(line number/character number will vary)

# Testing

I ran the tests, which all pass now that all `include` attributes have been removed from the XML files they use.

After I made the schema change, but before I changed the XML files used by tests, nearly all the tests failed.  This shows the `include` attribute will not be permitted in the next version of OpenMalaria.

While I determined from inspecting the codebase that the `include` attribute is not used, I ran some scenarios to confirm this out of caution.

I ran:

- one scenario with `include="0"` on all `<parameter>` elements,
- one scenario with `include="1"` on all `<parameter>` elements, and
- one scenario with no `include` attribute specified on any `<parameter>` elements.

All three scenarios resulted in identical output files, which confirms the attribute is unused.

### DIFF
--- a/schema/scenario.xsd
+++ b/schema/scenario.xsd
@@ -191,15 +191,6 @@ Licence: GNU General Public Licence version 2 or later (see COPYING) -->
         <xs:appinfo>units:Number;min:0;name:Parameter value;sweepable:true;</xs:appinfo>
       </xs:annotation>
     </xs:attribute>
-    <xs:attribute name="include" type="xs:boolean" use="optional">
-      <xs:annotation>
-        <xs:documentation>
-          True if parameter is to be sampled in optimization
-          runs. Not used in simulator app.
-        </xs:documentation>
-        <xs:appinfo>units:Number;min:0;max:1;name:Sampling indicator;exposed:false;</xs:appinfo>
-      </xs:annotation>
-    </xs:attribute>
   </xs:complexType>
   <xs:complexType name="Parameters">
     <xs:sequence>
--- a/test/scenario1.xml
+++ b/test/scenario1.xml
@@ -1102,36 +1102,36 @@
       </availabilityToMosquitoes>
     </human>
     <parameters interval="5" iseed="1" latentp="15d">
-      <parameter include="false" name="'-ln(1-Sinf)'" number="1" value="0.0507360000000001"/>
-      <parameter include="false" name="Estar" number="2" value="0.03247"/>
-      <parameter include="false" name="Simm" number="3" value="0.138161050830301"/>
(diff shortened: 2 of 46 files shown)


b-s-code:
@acavelan could I please get your review?  The diff is pretty cumbersome to look at, my apologies.

@melissapenny FYI

acavelan:
All good, I think this is necessary as many users have been wondering about use for these `include` parameters.

acavelan (approved):

b-s-code:
@acavelan Thanks for reviewing.

Something to note is that I spoke briefly with @melissapenny about this change; there was historically _another_ use for the `include` attribute that I was not aware of.

My current thinking is that if something similar is required by an external tool again in future, I'd be more likely to support metadata inside XML comments for that use case, or any other solution that allows things to remain simpler for the schema and other users.

b-s-code:
Double checked with @melissapenny - this is okay to merge 🟢

b-s-code:
@nakul7 FYI (just tagging you so you're aware of this change.)

PR #414, opened 2025-05-14
Elimination model update

acavelan:
Follow up of #402 #407 

Outputs related to totalPatentInfs_{introduced,indigenous} were 0 if totalPatentInfs was not also declared in the monitoring section. Each output can now be used independently of the others, as expected.

A new test scenario which uses all the new elimination outputs for imported, introduced and indigenous infections has been added.

## Breaking change

The measure nNewInfections now supports imported, introduced and indigenous infections. This is a breaking change as previous versions did not report imported infections through nNewInfections. The two tests that are using both nNewInfections and imported infections have been updated.

Imported infections are deployed at year 5 in the example below. We can then track the number of new imported, introduced and indigenous infections. 

Figure_1


b-s-code (commented):
Nice to see a new test added to cover this!

I left some initial remarks here but it looks like this PR depends on https://github.com/SwissTPH/openmalaria/pull/415 - so I'll probably delay final review of this PR until that one's done.

nakul7 (approved):
Approved as discussed on Tuesday.

test/scenarioInfectionOrigin.xml
```diff
@@ -0,0 +1,270 @@
...
+        </diagnostic>
+    </diagnostics>
+
+    <!-- Here you can specify additional model parameters, or model variants. -->
+    <model>
+        <ModelOptions>
+            <option name="INNATE_MAX_DENS" value="false"/> <!-- MANDATORY WITH THE BASE MODEL DO NOT REMOVE -->
+            <option name="INDIRECT_MORTALITY_FIX" value="false"/> <!-- MANDATORY WITH THE BASE MODEL DO NOT REMOVE -->
+            <option name="HEALTH_SYSTEM_MEMORY_FIX" value="false"/> <!-- MANDATORY WITH THE BASE MODEL DO NOT REMOVE -->
+        </ModelOptions>
```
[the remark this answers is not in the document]
acavelan: I just updated the test and realized it was missing.

b-s-code (approved):
Looks good to me.

acavelan:
Added imported, introduced and indigenous outputs to nNewInfections. First comment has been updated.

model/mon/OutputMeasures.h
```diff
@@ -289,8 +289,11 @@ void defineOutMeasures(){
     namedOutMeasures["Clinical_HospitalFirstDayDeaths"] =
         OutMeasure::humanAC( 42, MHO_HOSPITAL_FIRST_DAY_DEATHS, false );
     /** The number of actual infections since the last survey. */
-    namedOutMeasures["nNewInfections"] =
-        OutMeasure::humanAC( 43, MHR_NEW_INFECTIONS, false );
+    namedOutMeasures["nNewInfections"] = OutMeasure::humanAC( 43, MHR_NEW_INFECTIONS, false );
+        namedOutMeasures["nNewInfections_Imported"] = OutMeasure::humanAC( 1043, MHR_NEW_INFECTIONS_IMPORTED, false );
+        namedOutMeasures["nNewInfections_Introduced"] = OutMeasure::humanAC( 2043, MHR_NEW_INFECTIONS_INTRODUCED, false );
+        namedOutMeasures["nNewInfections_Indigenous"] = OutMeasure::humanAC( 3043, MHR_NEW_INFECTIONS_INDIGENOUS, false );
+
```
b-s-code, line 296: Reading the comment at the start of this method, it seems like the original intention for how the numerical output IDs would be organised in this method was:

1. they'd be kept in a strictly ascending order in source code, and
2. a developer adding a new output would add it near the end of the method and use the smallest possible unused numerical ID for their new output.

Clearly though, we do want to define things near each other sometimes (e.g. for readability, to demonstrate that the measures are related, etc.).

This conflicts with the original policy.

I think we should:

1. stop worrying about keeping numerical IDs in a sorted order altogether,
2. remove the comment stating the current policy for this method,
3. add a new comment stating "don't ever make an existing numerical identifier point to a new/different output measure because this would violate users' expectations"  (I think the most important thing is that the existing mapping between numerical IDs and output measures should only ever be "added to"), and
4. make it fail to compile if two different `OutMeasure`s are constructed with the same ID.

Regarding the last point, it appears that, currently, if 2043 was accidentally re-used like this, the code compiles happily.

```
        namedOutMeasures["nNewInfections_Introduced"] = OutMeasure::humanAC( 2043, MHR_NEW_INFECTIONS_INTRODUCED, false );
        namedOutMeasures["nNewInfections_Indigenous"] = OutMeasure::humanAC( 2043, MHR_NEW_INFECTIONS_INDIGENOUS, false );
```

I'm not suggesting any of this needs to happen as part of this PR though.

What do you think?

model/Host/ImportedInfections.h
```diff
@@ -116,7 +116,9 @@ namespace Host {
             if( rateNow > 0.0 ){
                 for(Human& human : population){
                     if(human.rng.bernoulli( rateNow )){
-                        human.withinHostModel->importInfection(human.rng, WithinHost::InfectionOrigin::Imported);
+                        human.withinHostModel->importInfection(human.rng);
+                        mon::reportEventMHI( mon::MHR_NEW_INFECTIONS, human, 1);
+                        mon::reportEventMHI( mon::MHR_NEW_INFECTIONS_IMPORTED, human, 1);
```
b-s-code, line 121: 👍👍

test/scenarioInfectionOrigin.xml
```diff
@@ -60,7 +60,11 @@
             <option name="inputEIR"/>                 <!-- id 35 -->
             <option name="simulatedEIR"/>             <!-- id 36 -->
             <option name="simulatedEIR_Introduced"/>  <!-- id 2036 -->
-            <option name="simulatedEIR_Indigenous"/>  <!-- id 2036 -->
+            <option name="simulatedEIR_Indigenous"/>  <!-- id 3036 -->
```
b-s-code, line 63: 👍👍👍

model/mon/OutputMeasures.h
b-s-code: Also it looks like the wiki docs don't have all the latest monitoring options/survey measures?

https://github.com/SwissTPH/openmalaria/wiki/MonitoringOptions#survey-measures

b-s-code (commented):
Overall new changes look good to me!

Looks like a couple things are due for some maintenance.  Keen to get your thoughts.

PR #415, opened 2025-05-14
Fix checkpointing

acavelan:
Checkpointing in tests was mistakenly disabled in [PR #339](https://github.com/SwissTPH/openmalaria/pull/339).

It has now been re-enabled, so all tests are run with checkpointing, ensuring that using checkpoints produces the same results as running without them.

This fixes checkpointing errors when running with non-default heterogeneity and when using the new elimination model.

### DIFF
--- a/model/Host/WithinHost/DescriptiveWithinHost.cpp
+++ b/model/Host/WithinHost/DescriptiveWithinHost.cpp
@@ -333,9 +333,8 @@ void DescriptiveWithinHostModel::checkpoint (istream& stream) {
 }
 void DescriptiveWithinHostModel::checkpoint (ostream& stream) {
     WHFalciparum::checkpoint (stream);
-    for(auto inf = infections.begin(); inf != infections.end(); ++inf){
+    for(DescriptiveInfection* inf : infections)
         *inf & stream;
-    }
 }
 
 char const*const not_impl = "feature not available with the \"descriptive\" within-host model";
--- a/model/Transmission/PerHost.h
+++ b/model/Transmission/PerHost.h
@@ -29,6 +29,7 @@
 #include "util/checkpoint_containers.h"
 
 #include "util/random.h"
+#include <fstream>
 
 namespace OM {
 namespace Transmission {
--- a/model/Transmission/PerHost.h
+++ b/model/Transmission/PerHost.h
@@ -29,7 +29,6 @@
 #include "util/checkpoint_containers.h"
 
 #include "util/random.h"
-#include <fstream>
 
 namespace OM {
 namespace Transmission {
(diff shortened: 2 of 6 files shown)


model/Host/WithinHost/DescriptiveWithinHost.cpp
```diff
@@ -333,9 +333,8 @@ void DescriptiveWithinHostModel::checkpoint (istream& stream) {
 }
 void DescriptiveWithinHostModel::checkpoint (ostream& stream) {
     WHFalciparum::checkpoint (stream);
-    for(auto inf = infections.begin(); inf != infections.end(); ++inf){
+    for(DescriptiveInfection* inf : infections)
```
b-s-code, line 336: Nice!

model/Transmission/PerHost.h
```diff
@@ -29,6 +29,7 @@
 #include "util/checkpoint_containers.h"
 
 #include "util/random.h"
+#include <fstream>
```
b-s-code, line 32: Is this include needed?

test/CMakeLists.txt
```diff
@@ -62,7 +62,7 @@ set (OM_BOXTEST_NC_NAMES)
 
 # Use python script.
 foreach (TEST_NAME ${OM_BOXTEST_NAMES})
-    add_test (${TEST_NAME} ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/run.py ${TEST_NAME})
+    add_test (${TEST_NAME} ${PYTHON_EXECUTABLE} ${CMAKE_CURRENT_BINARY_DIR}/run.py ${TEST_NAME} -- --checkpoint-stop)
```
b-s-code, line 65: 👍👍👍

model/Transmission/TransmissionModel.h
```diff
@@ -384,9 +384,11 @@ class TransmissionModel
...
         _annualAverageKappa &stream;
         _sumAnnualKappa &stream;
-        // tsAdultEIR &stream;
-        // surveyInputEIR &stream;
-        // surveySimulatedEIR &stream;
+        tsAdultEIR &stream;
+        surveyInputEIR &stream;
+        surveySimulatedEIR_i &stream;
+        surveySimulatedEIR_l &stream;
+        surveySimulatedEIR &stream;
```
b-s-code, line 391: I can understand why we want

```
        surveySimulatedEIR_i &stream;
        surveySimulatedEIR_l &stream;
```

given we have `surveySimulatedEIR &stream;`, in the context of recent changes 👍

Were the following bits commented out in https://github.com/SwissTPH/openmalaria/pull/384/ erroneously, or for a reason that no longer holds?

```
tsAdultEIR &stream;
surveyInputEIR &stream;
surveySimulatedEIR &stream;
```

acavelan: It was commented erroneously and it didn't trigger an error in the tests because checkpoint tests were disabled. These are needed!

b-s-code (approved):
All looks good now.

PR #417, opened 2025-05-20
add detection of possible duplicate OutId

acavelan:
Follow up on: https://github.com/SwissTPH/openmalaria/pull/414#discussion_r2096872718

This will throw an error at runtime if duplicates exist.

This only affects developers but this will help avoid having duplicate outIds by mistake.

It would be better to have a detection at compile time, and while I'm sure it's possible with some clever templates, I'm not sure it's needed. The current code will throw an error and fail all the tests as expected, even if the outputMeasures are not used.


b-s-code (approved):
I think this is a great solution.

PR #419, opened 2025-06-09
Support named models

b-s-code:
## Problem

For a user, creating a scenario using the base model is more difficult and error-prone than strictly required.

When I say the "base model" I refer to the collection of [model options specified here](https://github.com/SwissTPH/openmalaria/blob/9679433fa472a44417dfe7a8facec79b80e1fc2f/util/example/example_scenario.xml#L273-L276) and [the parameter values specified here](https://github.com/SwissTPH/openmalaria/blob/9679433fa472a44417dfe7a8facec79b80e1fc2f/util/example/example_scenario.xml#L311-L342).

## Solution

This PR makes it possible for users to write a model name (only the name "base" is supported in this PR) instead of the user explicitly writing out/copying collections of model options and parameter values.

This is accomplished by changing the schema to accept a new `ModelName` element.  E.g.

```
...
    <model>
        <ModelName name="base"/>
        ...
        <!-- User can omit <ModelOptions> element. -->
        ...
        <!-- User can omit <parameters> element. -->
    </model>
...
```

Here is how the new `ModelName` element works in conjunction with the existing `ModelOptions` and `Parameters` elements.

| `<ModelName>` included? | `<ModelOptions>` included? | `<parameters>` included? | Supported?     | What happens? |
|---------------|------------------|----------------|----------------|---------------|
| ❌        | ❌           | ❌         | ❌             | OM crashes with error message.          |
| ❌        | ❌           | ✅        | ❌             | OM crashes with error message.          |
| ❌        | ✅          | ❌         | ❌             | OM crashes with error message.          |
| ❌        | ✅          | ✅        | ✅            | Simulation runs as before.  |
| ✅       | ❌           | ❌         | ✅            | Model options/params from named model are used. |
| ✅       | ❌           | ✅        | ✅            | Model options/params from named model are used, but explicitly specified entries in `<parameters>` override those. |
| ✅       | ✅          | ❌         | ✅            | Model options/params from named model are used, but explicitly specified entries in `<ModelOptions>` override those. |
| ✅       | ✅          | ✅        | ✅            | Model options/params from named model are used, but explicitly specified entries in `<ModelOptions>`/`<parameters>` override those. |

## Testing

- Added new tests to cover supported combinations of ModelName/Parameters/ModelOptions.  Tried to make these not run too long.
  - Did not add any new automated tests to verify the behaviour resulting from unsupported combinations (i.e. no ModelName and one/both of Parameters/ModelOptions also missing).  Would require [something like googletest's "death tests"](https://github.com/google/googletest/blob/main/docs/advanced.md#death-tests) - I'm not sure if/how this may be achievable with the current testing framework.

## Other notes

This PR also includes various housekeeping/refactoring/extension...

- The design of this change makes it easy enough to extend OpenMalaria with new named models in future.
- I chose to hardcode the values to use for the base model.  In this case, I think hardcoding the values is preferable to increasing the number of files that OpenMalaria reads, and hardcoding couples named-model versions to software versions (which I think is probably a good thing).
- I refactored the `Parameters` class to modernise it (getting rid of a raw enum etc.) and reflect the fact that each parameter ultimately is associated with a 3-tuple (integer ID, parameter name, parameter value).
- Fixed reference to specific schema version in `test/scenarioVivax.xml`.
- Removed default time units for the `latentp` attribute.  While I thought the docs provided reasonably clear info, I had noticed users expressed uncertainty about the previous behaviour.  The attribute now requires units to be specified explicitly (both days and timesteps are still accepted).    I updated test inputs to specify time units for `latentp` explicitly.
- Split `<computationParameters>` from `<parameters>`.  `<parameters>` should be able to be omitted, but we generally want `iseed` to be configurable without having to specify `<parameters>` explicitly.
- UnittestUtil - I really just made enough changes here so that old cxxtest unit tests would still run.
- Added a class XMLChecker to check whether the scenario contents are appropriate before beginning the simulation.  This makes it possible to:
  - check things that are difficult or impossible to enforce in the schema itself,
  - to check those things early, instead of failing once the simulation has been running for a while, and
  - to provide relevant error messages.
 - I manually updated a few copyright headers in this PR before learning the script `https://github.com/SwissTPH/openmalaria/blob/main/util/update-header.py` exists for this.  I don't plan to update the script or run it within this PR since this PR is already big enough.

## Documentation

- I grepped the OM wiki repo for `ModelOptions`, `model`, `param` and `parameters` XML tags with `rg -l -i "<model|<param"` and inspected relevant pages.
- Pushed a branch `bs_modelname` with planned wiki docs changes, related to this PR.  GitHub can't do PRs for wiki repos, it seems.
- Added proposed release notes for this change to the aforementioned branch.

It's too early to merge changes to docs, since this feature in this PR is not yet released.  So instead I raised an issue to represent the task of merging the docs changes etc.: https://github.com/SwissTPH/openmalaria/issues/420.


b-s-code:
@acavelan can I please get a review?

@melissapenny @nakul7 FYI

acavelan (commented):
It’s a big step in the right direction, in my opinion! It simplifies the XML and should help avoid many costly mistakes.

The logic to override some parameters if needed looks good to me and it's quite intuitive.

Having the seed parameter outside make sense as well. 

The XMLChecker looks good!

Some comments I have:
- I like Parameters:: better than ParameterName::, it's a bit shorter (but I leave it up to you)
- latentp and interval should probably both be provided with the rest of the model parameters. No one should be able to change it unless they really know what they are doing.

b-s-code:
Thanks for the initial review @acavelan .

>     * I like Parameters:: better than ParameterName::, it's a bit shorter (but I leave it up to you)

Previously, it was possible to refer to a parameter with something like `Parameters::E_STAR` when a raw enum was used:

```cpp
class Parameters
{
public:
    enum Parameter // Raw enum means that the identifiers contained here are accessible in parent scope.
    {
        NEG_LOG_ONE_MINUS_SINF,
        E_STAR,
        // ...
    };
};
```

But stopped being possible when I changed the raw enum to an enum class.  With the following code, referring to `Parameters::E_STAR`  no longer compiles.  Instead, `Parameters::Parameter::E_STAR` became the necessary way to refer to the entries of the enum.

```cpp
class Parameters
{
public:
    enum class Parameter // Usage of enum class here removes namespace promotion.
    {
        NEG_LOG_ONE_MINUS_SINF,
        E_STAR,
        // ...
    };
};
```

As such, I moved the enum class outside the Parameters class, so that it became possible to refer to a parameter with something shorter like `ParameterName::E_STAR`.

I'd prefer to keep using an enum class, rather than a raw enum.

I can't rename `ParameterName` to `Parameters` since that conflicts with the name of the `Parameters` class.  But I can call it just `Parameter`, if you're happy with that?  That's actually what it was originally called.

>     * latentp and interval should probably both be provided with the rest of the model parameters. No one should be able to change it unless they really know what they are doing.

Sounds reasonable to me but we'll need to consider to implement this, since the other model parameters are all `double`s and both `interval` and `latentp` are really integers with time units.  A few options come to mind, I'll have a think.  Any ideas are welcome.

model/Host/WithinHost/WHVivax.cpp
```diff
@@ -580,12 +581,22 @@ double WHVivax::getCumulative_h() const{ throw TRACED_EXCEPTION( not_impl, util:
...
 
 void WHVivax::init( const OM::Parameters& parameters, const scnXml::Model& model ){
-    try{
-        //NOTE: if XSD is changed, this should not have a default unit
-        latentP = UnitParse::readShortDuration( model.getParameters().getLatentp(), UnitParse::STEPS );
-    }catch( const util::format_error& e ){
-        throw util::xml_scenario_error( string("model/parameters/latentP: ").append(e.message()) );
+
+    // Default value to use if no <parameters> element is present in XML.
+    latentP = UnitParse::readShortDuration("15d", UnitParse::NONE);
```
acavelan, line 586: Also here

b-s-code:
@acavelan I've pushed some changes, and added some more remarks.  Please take another look and let me know your thoughts.

At the moment, I'm leaning towards addressing most of the outstanding items in a follow up PR.  Happy to include outstanding here changes instead though, if desired.

Issue #420, opened 2025-06-13
Update example scenario and docs to go with release of named models feature.


b-s-code:
https://github.com/SwissTPH/openmalaria/pull/419 adds support for named models.

Wiki docs require updates to reflect the changes, and so does the example scenario XML file.

These changes should not be merged to main branch in the OM repo and wiki repo until just before/after the feature is released.

This issue is to track that the changes do get made.

b-s-code:
I pushed some wiki docs changes to a branch `bs_modelname`.

Still need to explain the named models feature/the base model itself in more more detail before merging.  Best place for this is probably under the "Specification of model variants" header in ModelWithinHost.md.

b-s-code:
All done.

PR #422, opened 2025-06-20
Add a default time step duration to the base model

b-s-code:
## Problem

The `interval` value in XML scenarios represents the *number of days per timestep*.

We'd like `interval` to be an attribute that:

- users do not *have to* set manually, in order to use the base model, but
- users *can* set manually *if they need to*.

## Solution

1. In terms of the schema, I moved the `interval` attribute from the `<computationParameters>` element back to the `<parameters>` element.  I did this because #419  made it possible to omit the `<parameters>` element when using the base model, and we can use that functionality to accomplish most of what's needed for the feature in this PR.
2. I added a value for the `interval` (in particular, a default of a **5 day time step**) to the base model.  I did this by modifying the `sim` class, which already manages the `interval` value.  This required:
    - adding a dependency of the `sim` class on the `ModelNameProvider` class, and
    - changing the order we use to initialise things in `main()`.

3. I updated unit test code and black box test inputs to accommodate the above changes.

Note: this PR does not make any changes related to how the `latentp` attribute is treated in the schema or handled in the C++ code.

## Testing

- [x] Existing tests work fine now input scenarios are migrated and unit tests code has been updated (no change was made to outputs of black box tests).

### Manual testing

- [x] Verified that a helpful error message is printed when specifying a model name other than "base".  `Error: Unrecognized model name: base2` (I chose "base2" as an arbitrary example of an invalid model name).
- [x] Confirmed that a user can override base model `interval` value with custom value in parameters block.  I tested this using `scenarioEmpirical.xml`.  That scenario requires `interval` to have a value 1, otherwise OpenMalaria exits and prints "Error: Model option CLINICAL_EVENT_SCHEDULER is only compatible with a 1-day time step.".  I added `<ModelName name="base"/>` to `scenarioEmpirical.xml`, and unlike manually setting `interval="5"`, OpenMalaria did not crash with that error message - instead it ran successfully.  This indicates the mechanism for overriding the base model's `interval` value is working.

## Documentation

I've got a branch of the wiki repo with documentation updates related to this change.

To managed the task of merging the wiki update at the appropriate time, I have an issue https://github.com/SwissTPH/openmalaria/issues/420 .

## Notes

A couple of the scenario files for black box test have a massive diff.  I think this is probably due to a change in line endings.  I can make a more surgical change to those files if desired, to ease burden of reviewing.

### DIFF
--- a/model/main.cpp
+++ b/model/main.cpp
@@ -149,22 +149,22 @@ int main(int argc, char* argv[])
 
         util::XMLChecker().PerformPostValidationChecks(*scenario);
 
-        sim::init(*scenario); // also reads survey dates
-
         // 1) elements with no dependencies on other elements initialised here:
-        util::ModelNameProvider modelNameProvider(scenario->getModel());
         WithinHost::Genotypes::init( *scenario );
-
         util::master_RNG.seed( scenario->getModel().getComputationParameters().getIseed(), 0 ); // Init RNG with Iseed
+        util::ModelNameProvider modelNameProvider(scenario->getModel());
 
         // 2) elements depending on only elements initialised in (1):
+        sim::init(*scenario, modelNameProvider); // Also reads survey dates.
+
+        // 3) elements depending on only elements initialised in (2).
         Parameters parameters( scenario->getModel().getParameters(), modelNameProvider ); // Depends on ModelNameProvider.
         util::ModelOptions::init( scenario->getModel().getModelOptions(), modelNameProvider ); // Depends on ModelNameProvider.
-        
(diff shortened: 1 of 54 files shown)


b-s-code:
@acavelan I mentioned I'd follow up with some changes for the `interval` and `latentp` attributes on OpenMalaria's model parameters.  This PR only addresses `interval`, I haven't dealt with `latentp` at all here.  Can I please get a review?

FYI @melissapenny @nakul7

acavelan (approved):

PR #425, opened 2025-07-02
refactor: Make parameter initialisation logic re-usable.

b-s-code:
## Problem

https://github.com/SwissTPH/openmalaria/pull/419 added "named models" (collections of model options and parameter values) as a feature.

To add further named models, it will be useful to re-use some of the parameter initialisation logic added in the above PR.

## Solution

1. Created a new method which encapsulates some the logic used to initialise each of the base models parameters.
2. Made the base model initialisation code use this new method.

## Testing

Automated tests all still pass.

### DIFF
--- a/model/Parameters.h
+++ b/model/Parameters.h
@@ -256,10 +256,11 @@ private:
     }
 
     /*
-    * Initializes some hardcoded values for some parameters, encapsulating the parameter values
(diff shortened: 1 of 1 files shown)


acavelan (approved):
It looks good to me!

PR #426, opened 2025-07-04
chore: Remove unnecessary enum.

b-s-code:
# Problem

At time of writing, OpenMalaria is built against the C++17 standard.

https://github.com/SwissTPH/openmalaria/blob/107f4de5d1639271d3994031ea4fccbcc5a8be59/CMakeLists.txt#L15

I tried to compile OpenMalaria with gcc using C++20 to get a feel for what would be involved in upgrading.

Many warnings resulted because arithmetic operations between floating-point types and enumeration types are deprecated in C++20 (at least w.r.t. gcc)

E.g.

```
warning: arithmetic between floating-point type ‘double’ and enumeration type ‘OM::SimData::<unnamed enum>’ is deprecated [-Wdeprecated-enum-flo
at-conversion]                                                                                                                                                                                 
   88 |     static inline double inYears(SimTime d) { return d * (1.0 / SimData::DAYS_IN_YEAR); }   
```

This PR addresses the source of many of those warnings.

This PR does not:

- address *all* issues relating to building with C++20 (in particular, there appear to be warnings when building unit test with C++20, which are not addressed here, and I've not tested any other compilers we support), or
- actually make the switch to C++20.

## Solution

1. Removed redundant definition of DAYS_IN_YEAR.  (It was defined twice).
2. Made usage sites of DAYS_IN_YEAR use the other, remaining definition instead.
3. Replaced the single-entry enum with a `static const int`.

## Testing

Automated tests still run and pass.

### DIFF
--- a/model/sim.h
+++ b/model/sim.h
@@ -46,8 +46,6 @@ inline int floorToInt( double x ){
 /// Internal use only
 // Required to avoid use-before-definition
 class SimData {
-    enum { DAYS_IN_YEAR = 365 };
-    
     static inline int interval;        // days per time step
     static inline size_t steps_per_year;
     static inline double years_per_step;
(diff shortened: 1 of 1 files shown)


b-s-code:
@acavelan could I please get a review?

FYI @melissapenny @nakul7

PR #430, opened 2025-07-11
fix: Rectify gh action's effect on gh release page text.

b-s-code:
# Problem

The description on the GitHub Release page for OpenMalaria was being overwritten with "OpenMalaria build for Windows".  This text is meant to pertain to the whole release - for all platforms.

## Solution

This is a band-aid.

This PR does not stop the release's description text being overwritten.

This PR makes the release's description text get overwritten with the desired text.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -35,7 +35,11 @@ artifacts:
 deploy:
   force_update: true
   provider: GitHub
-  description: "OpenMalaria build for Windows"
+  description: |
+    Release of OpenMalaria.
+    
+    Updates: https://github.com/SwissTPH/openmalaria/wiki/SchemaUpdateGuide
+    Changelog: https://github.com/SwissTPH/openmalaria/wiki/Changelog
   auth_token:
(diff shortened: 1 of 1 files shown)


acavelan:
It needs testing!

b-s-code:
I've pushed a tag in my fork to test these changes.
https://github.com/b-s-code/openmalaria/releases/tag/schema-test_new_build

The last couple links in the release text are broken, which is fair enough because I just made a dummy tag in my fork.

Example scenario: https://github.com/SwissTPH/openmalaria/blob/schema-test_new_build/util/example/example_scenario.xml
XSD documentation: https://swisstph.github.io/openmalaria/schema-test_new_build.html

I'm waiting to see if the release text is preserved after the appveyor build finishes.

Then I'll follow up.

b-s-code:
Following up: the AppVeyor build didn't overwrite the release text which is good.  But now the Windows build doesn't appear at all on the release page, even though it succeeded.  I'll have to look into it.

acavelan (approved):

Issue #436, opened 2025-09-01
macos CI build fail with errors related to cmake and xsd.


b-s-code:
Originally noticed when syncing my fork of OpenMalaria with upstream.

Managed to reproduce in primary OpenMalaria repo.  https://github.com/SwissTPH/openmalaria/pull/435

As far as I can tell, this is not a result of any recent OpenMalaria code changes.

PR #437, opened 2025-09-01
chore: Fix macOS CI builds.

b-s-code:
# Problem

Recently, macOS CI builds started failing, for all macOS versions that we target.

See issue https://github.com/SwissTPH/openmalaria/issues/436

This seems likely to be caused by a recent change to GitHub's runner images for macOS https://github.com/actions/runner-images/pull/12791 and there is an open issue related to cmake failing to install on actions that use those images https://github.com/actions/runner-images/issues/12912

## Solution

I was able to take the blame off any recent OpenMalaria changes by reproducing the issue with the commit tagged `schema-48.0`, which we know to have passed CI.

Updating the relevant GitHub Actions workflow config file to no longer install certain packages stops the relevant builds failing.

To some extent, this solution is a workaround, and I think it's better to have this workaround now than have failing CI until GitHub might fix the issue.

## Testing

As can be seen in the CI checks on this PR, the relevant builds now pass (green ticks).

Looking more closely at the logs for those builds, it is clear that there are still some warnings and one error.  I suggest we live with them.  They do not indicate any relevant defect in the macOS environments we're using for testing OpenMalaria.

**macOS 14 and macOS 15**

```
Warning: Already linked: /opt/homebrew/Cellar/xsd/4.2.0
To relink, run:
  brew unlink xsd && brew link xsd
```

**macOS 13**

```
Error: The `brew link` step did not complete successfully
The formula built, but is not symlinked into /usr/local
Could not symlink share/man/man1/xsd.1
Target /usr/local/share/man/man1/xsd.1
already exists. You may want to remove it:
  rm '/usr/local/share/man/man1/xsd.1'

To force the link and overwrite all conflicting files:
  brew link --overwrite xsd

To list all files that would be deleted:
  brew link --overwrite xsd --dry-run

Possible conflicting files are:
/usr/local/share/man/man1/xsd.1 -> /Library/Frameworks/Mono.framework/Versions/Current/share/man/man1/xsd.1
```

### DIFF
--- a/.github/workflows/build-and-deploy.yml
+++ b/.github/workflows/build-and-deploy.yml
@@ -72,7 +72,7 @@ jobs:
       - name: Install - MacOS
         if: startsWith(matrix.os,'macos')
         run: |
-          HOMEBREW_NO_AUTO_UPDATE=1 brew install git coreutils cmake gcc gsl xerces-c xsd || true
+          HOMEBREW_NO_AUTO_UPDATE=1 brew install coreutils gsl xerces-c xsd || true
           HOMEBREW_NO_AUTO_UPDATE=1 brew link --overwrite xsd
       
       - name: Infos


b-s-code:
So GitHub's bug fix didn't resolve failures fully for macOS 14.

Failures for macOS 13 and 15 were resolved by their bug fix though.

It is possible to take a slightly smaller changeset than this PR includes (i.e. can just remove only cmake installation statement).

Either way we get errors/warnings in the CI logs but green ticks, as long as build/tests pass.

Given that GitHub marked the relevant bug as "fixed" now, I propose we merge this PR's changes.

Thoughts @acavelan ?

acavelan:
I believe the error you quoted indicates that the xsd dependency is already installed, too. So probably we can remove this one as well.

acavelan (approved):

b-s-code:
Removing xsd breaks the relevant builds, so I'll leave it in.

```
 Building...
/opt/homebrew/bin/cmake
-- The CXX compiler identification is AppleClang 17.0.0.17000013
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
CMake Error at cmake/FindXSD.cmake:42 (message):
  Unable to find xsd include files (xsd/cxx/parser/elements.hxx)
Call Stack (most recent call first):
  CMakeLists.txt:129 (find_package)
```

Thanks for reviewing!

PR #438, opened 2025-09-08
refactor: Remove unsupported growthRate for demography.

b-s-code:
# Problem

While reading some age structure code, I encountered an unexpected XML attribute.

This attribute is related to population growth.

For at least 10 years, OpenMalaria has just thrown an exception if this attribute is used.

## Solution

This PR removes schema-level support for the XML attribute and deletes code related to it.

## Testing

(macOS CI fails but this is addressed in a separate PR https://github.com/SwissTPH/openmalaria/pull/437)

Nonzero values for the growthRate attribute were not supported before this PR anyway.

**Before:**

When one added growthRate="0.1" to the <demography> node in the example, XML, the output when running OpenMalaria is:
Error: Population growth rate provided.
In: example_scenario.xml

**After:**

Now, if one adds growthRate="0.1" to the <demography> node in the example, XML, the output when running OpenMalaria is:
:29:83 error: attribute 'growthRate' is not declared for element 'demography'

## Documentation

Wiki docs are already  consistent with the behaviour before and after this PR, no update is needed.

https://github.com/SwissTPH/openmalaria/wiki/ModelDemography
 
> The simulations also run with constant population sizes

## Notes

I chose to leave a variable called `rho` instead of just hardcoding a value of 0 everywhere it's used because that seems more readable to me.

### DIFF
--- a/model/Population.cpp
+++ b/model/Population.cpp
@@ -81,8 +81,6 @@ void Population::update()
 {
     //NOTE: other parts of code are not set up to handle changing population size. Also
     // size is assumed to be the _actual and exact_ population size by other code.
-    //targetPop is the population size at time t allowing population growth
-    //int targetPop = (int) (size * exp( AgeStructure::rho * sim::ts1().inSteps() ));
     int cumPop = 0;
 
     for (auto it = humans.begin(); it != humans.end();) {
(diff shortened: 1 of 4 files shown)


b-s-code:
@acavelan could I please get a review?

@melissapenny @nakul7 FYI

acavelan (approved):

PR #448, opened 2026-01-21
Add emergence rate root finding calculation

acavelan:
Follow up on #279 

## Problem

The existing method to calculate the emergence rate (Nv0) consists in scaling and shifting the initial Sv vector (the population of infectious mosquitoes initialized from the seasonal profile) and comparing the resulting SimulatedEIR to the InputEIR after 5 years of additional simulation time. This process can be repeated up to 30 times (up to 150 years of additional simulation) to reach an acceptable error threshold, which is currently defined as being within 10% of the InputEIR. This is done for each species, sequentially.

This is a long process and 10% is a relatively large error threshold. This is because Sv -> SimulatedEIR depends on many factors including changes in immunity and human infectivity, which take time to stabilize and create a feedback loop with the SimulatedEIR.

For extreme seasonal profiles and/or low EIR values, this method sometimes fails to reach the minimum error threshold even after 30 iterations.

## Solution 

Calculate the emergence rate according to the system of equations described in https://doi.org/10.1007/s11538-011-9710-0, by Nakul Chitnis. We use a root-finding algorithm to compute the unique solution in one time-step.

Because the unique solution is calculated from the equations and not inferred, the result is that the difference between the SimulatedEIR and the InputEIR is much smaller on average. This is especially noticeable for low annual EIR values. 

The root-finding algorithm is also extremely fast and we no longer need to rely on the the long iterative process described above.

### Note 

We currently use a root-finding algorithm because it is fast and well-understood. We note that a linear solver has been tested and produces the same results on all tested scenarios so far,and is faster. However, the derivations to arrive at the linear formulation are more complex and need a formal proof. Both the root-finding algorithm and the linear solver rely on the GSL library, which is already a dependency of OpenMalaria.

## New Model Option

USE_EXACT_NV0_SOLVER is used for choosing the Nv0 emergence-rate initialization method. The option defaults to false. The default remains the EmergenceRateAdaptiveFitter for now.

## Tests

A new test scenario has been added to test the new solver.

## Results

The new solver can sometimes produce negative emergence rate values when the EIR seasonal profile has a sharp peak (fast increase, fast decrease). While the old method would sometimes still run and produce questionable results, the new method will fail explicitly and immediately. This will result in more failures than before but it also ensures that only scenarios where the SimulatedEIR matches the InputEIR very well will work, and it is up to the user to smooth the EIR seasonal profile. 

If the emergence rate vector contains negative values, the user will receive the following error:
`Error: The exact emergence solver produced negative emergence rates. With the current mosquito mortality/survival parameters, the requested EIR likely falls faster than the infectious mosquito population can decline biologically; matching it exactly would require negative emergence. Recommended action: smooth the EIR seasonality. Alternatively, remove <option name="USE_EXACT_NV0_SOLVER" value="true"/> to use the legacy adaptive fitter. This may allow the simulation to run, but the simulated EIR may deviate more from the input EIR.
`

## Example results

Figure_1

Figure_2

## Heatmap

image

image


b-s-code:
Is the top heatmap showing error when using legacy adaptive fitter?

And bottom showing error using new root finding algo?

The heatmaps have same titles.

b-s-code:
Will the plots on this wiki page need updating after the PR is merged?

https://github.com/OpenMalaria-Org/openmalaria/wiki/ScenarioTransmission#limits-of-the-model

acavelan:
> Is the top heatmap showing error when using legacy adaptive fitter?
> 
> And bottom showing error using new root finding algo?
> 
> The heatmaps have same titles.

Yes and yes.

acavelan:
> Will the plots on this wiki page need updating after the PR is merged?
> 
> https://github.com/OpenMalaria-Org/openmalaria/wiki/ScenarioTransmission#limits-of-the-model

Yes.

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+
+#include "CalcNvO.h"
+#include "util/errors.h"
+
+// Regularized smoothness solve for Nv0:
+//   x = argmin ||J x - S||_2^2 + lambda * ||D2 x||_2^2
+// where D2 is the *periodic* second-difference operator.
+// Solves (J^T J + lambda * (D2^T D2)) x = J^T S via Cholesky (SPD).
+//
+// Requires GSL: <gsl/gsl_blas.h>, <gsl/gsl_linalg.h>
```
b-s-code: Seems to require more gsl headers than just these. Easier to delete the comment?

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+#include <vector>
+#include <algorithm>
+#include <cmath>
+#include <gsl/gsl_matrix.h>
+#include <gsl/gsl_vector.h>
+#include <gsl/gsl_linalg.h>
+#include <gsl/gsl_blas.h>
+
+static inline int modp(int i, int n) {
+    int r = i % n;
```
b-s-code, line 42: const

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+// x must be allocated length n.
+static int solve_Nv0_smooth_reg(
+    gsl_vector* x,
+    const gsl_matrix* J,
+    const gsl_vector* S,
+    double lambdaD2,
+    double lambdaRidge
+) {
+    const int n = static_cast<int>(J->size1);
+    if (J->size2 != (size_t)n || S->size != (size_t)n || x->size != (size_t)n) {
```
b-s-code: Might be more concise to just make n size_t above.

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+        gsl_matrix_set(A, i, im1, gsl_matrix_get(A, i, im1) + lambda * (-4.0));
+        gsl_matrix_set(A, i, ip1, gsl_matrix_get(A, i, ip1) + lambda * (-4.0));
+        // Second off-diagonals
+        gsl_matrix_set(A, i, im2, gsl_matrix_get(A, i, im2) + lambda * 1.0);
+        gsl_matrix_set(A, i, ip2, gsl_matrix_get(A, i, ip2) + lambda * 1.0);
+    }
+}
+
+// Solves for x (length n):
+//   (J^T J + lambdaRidge * I + lambdaD2 * D2^T D2) x = J^T S
```
b-s-code: In this comment and some others below, it might be clearer if a multiplication operator appeared between matrices.

E.g. `J^T J` just uses concatenation to imply multiplication (I'm assuming) but the following term in the sum explicitly shows a multiplication operator.

Or not, if wanting to distinguish scalar/matrix operations.

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+static double compute_lambdaSmooth_from_J(
+    const gsl_matrix* J,
+    double cSmooth,
+    bool print_diag = true)
+{
+    const double n = static_cast<double>(J->size1); // assume square n×n
+    const double JF2 = frob_norm_sq(J);
+    const double D2F2 = 6.0 * n;   // periodic second difference operator
+
+    double lambda = cSmooth * (JF2 / D2F2);
```
b-s-code: const

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+
+static void CalcXP_optimized(gsl_vector** xp, const std::vector<gsl_matrix*> &Upsilon,
+		gsl_vector** Lambda, gsl_matrix* inv1Xtp, int eta, int thetap) {
+	gsl_vector* vtemp = gsl_vector_calloc(eta);
+	gsl_matrix* mtemp = gsl_matrix_calloc(eta, eta);
+	gsl_matrix* tail = gsl_matrix_calloc(eta, eta);
+	gsl_matrix* mmul = gsl_matrix_calloc(eta, eta);
+	gsl_vector* x0p = gsl_vector_calloc(eta);
+
+	// Performance patch from 7b6a2b91: compute x0p with a backward tail-product
```
b-s-code: Can't find any commit 7b6a2b91... I assume this refers to a commit in the original codebase this file is from?

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+		gsl_vector_free(Lambda[i]);
+		gsl_vector_free(xp[i]);
+	}
+
+	free(Lambda);
+	free(xp);
+	gsl_vector_free(SvfromNv0);
+}
+
+static int CalcSvDiff_rf(const gsl_vector* x, void* p, gsl_vector* f) {
```
b-s-code, line 164: Is there any strong reason to use void* here?

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+
+using namespace std;
+
+/* calcInitMosqEmergeRate() calculates the mosquito emergence rate given
+ * all other parameters.
+ *
+ * We use a periodic version of the model described in "A Mathematical Model 
+ * for the Dynamics of Malaria in Mosquitoes Feeding on a Heteregeneous Host
+ * Population". The periodic model still needs to be written as a paper. We will
+ * change these comments to refer to the approprirate paper when it is ready.
```
b-s-code: Should this comment now reference https://link.springer.com/article/10.1007/s11538-011-9710-0 ?

"A Periodically-Forced Mathematical Model for the Seasonal Dynamics of Malaria in Mosquitoes"

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+	const std::vector<double> &PBi, // $P_{B_i}$: mosqProbBiting (length n)
+	const std::vector<double> &PCi, // $P_{C_i}$: mosqProbFindRestSite (length n)
+	const std::vector<double> &PDi, // $P_{D_i}$: mosqProbResting (length n)
+	const std::vector<double> &PEi, // $P_{E_i}$: mosqProbOvipositing -- assumed not affected by nhh
+	const std::vector<double> KviInit, // Kvi (size n * thetap)
+	const std::vector<double> SvInit // Sv (length n)
+){
+    /* Note that from here on we use the notation from "A Mathematical Model for the
+	 * Dynamics of Malaria in Mosquitoes Feeding on a Heterogeneous Host Population",
+	 * and (the publication with the periodic model - yet to be written).
```
b-s-code: In need of an update?

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+	 * to the equations, as they will be written in the code. Since the equations are not
+	 * obvious in any case, anyone who wants to go through this code, will need to go 
+	 * through the paper, so I think that will be ok.
+	 *
+	 * There are also a number of variables defined that are difficult to describe
+	 * physically which we use in intermediate equations. We try to give names that
+	 * we use in the papers referenced above. 
+	 *
+	 *  - Any complaints about this notation (or anything else in general) can be directed 
+	 * to itsupport-sti@stimail.ch
```
b-s-code: Still relevant?

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+
+	// If the spectral radius of Xtp is greater than or equal to 1, then
+	// we are not guaranteed the existence of a unique globally asymptotically
+	// stable periodic orbit; thus it does not make sense to try to match the EIR
+	// for this periodic orbit. 
+	//
+	// For this model, all the eigenvalues should be in the unit circle. However,
+	// as we cannot show that analytically, we need to check it numerically.
+	if(srXtp >= 1.0){
+		// Throw an error.
```
b-s-code: Nothing thrown?

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+		throw ::OM::util::base_exception(
+			"The exact emergence solver produced negative emergence rates. "
+			"With the current mosquito mortality/survival parameters, the requested EIR likely "
+			"falls faster than the infectious mosquito population can decline biologically; "
+			"matching it exactly would require negative emergence. "
+			"Recommended action: smooth the EIR seasonality. "
+			"Alternatively, set <option name=\"USE_EXACT_NV0_SOLVER\" value=\"false\"/> "
+			"to use the legacy adaptive fitter. This may allow the simulation to run, "
+			"but be aware that the simulated EIR may deviate more from the input EIR.",
+			::OM::util::Error::VectorFitting);
```
b-s-code, line 487: Looks good.

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+	// period. 
+	// This is the sum of j from 0 to k_+ in (2.3c).
+	double sumkplus = 0; 
+
+	// This is an array of the sums from 0 to k_{l+} in (2.3c).
+	// Note that sumklplus here is defined as sumlv in MATLAB.
+	double* sumklplus = (double *)malloc((tau-1)*sizeof(double));
+
+	// Calculate probabilities of mosquito surviving the extrinsic
+	// incubation period.]
```
b-s-code: A stray "]"?

model/Transmission/Anopheles/CalcNvO.cpp
```diff
@@ -0,0 +1,1099 @@
...
+			  int tau, double PA, double Pdf){
+	int kplus = thetas / tau - 1; // $k_+$ in model.
+
+	// Evaluate sumkplus
+	sumkplus = 0.;
+	for (int j=0; j <= kplus; j++){
+		double tempbin = gsl_sf_choose(thetas - (j + 1) * tau + j, j); // binomial
+		double temppap = pow(PA,thetas-(j+1)*tau);
+		double temppdfp = pow(Pdf,j);
+		double temp = tempbin*temppap*temppdfp;
```
b-s-code: Const?

model/Transmission/Anopheles/EmergenceRateEstimator.h
```diff
@@ -98,16 +100,25 @@ inline double findAngle(const double EIRRotageAngle, const vector<double> & FSCo
...
 }
 
-class AnophelesModelFitter
+class EmergenceRateEstimator
+{
+public:
+    virtual ~EmergenceRateEstimator() = default;
+
+    // Interface: returns true if further fitting iterations are needed.
+    virtual bool estimate(AnophelesModel& m, const vector<double> &laggedKappa, double meanAvail) = 0;
```
b-s-code: Would prefer std:: prefix for vector etc.

Also elsewhere in this file.

model/Transmission/Anopheles/EmergenceRateEstimator.h
```diff
@@ -98,16 +100,25 @@ inline double findAngle(const double EIRRotageAngle, const vector<double> & FSCo
     return minAngle;
 }
 
-class AnophelesModelFitter
+class EmergenceRateEstimator
+{
+public:
+    virtual ~EmergenceRateEstimator() = default;
+
+    // Interface: returns true if further fitting iterations are needed.
```
b-s-code, line 108: Might have missed something but looks like the implementation only ever returns false?

model/Transmission/transmission.h
```diff
@@ -343,10 +345,14 @@ inline VectorModel *createVectorModel(const scnXml::Entomology &entoData, int po
         PerHostAnophParams::init(anoph.getMosq());
 
         Anopheles::AnophelesModel *anophModel = createAnophelesModel(i, anoph, initialisationEIR, populationSize, interventionMode);
-        Anopheles::AnophelesModelFitter *fitter = new Anopheles::AnophelesModelFitter(*anophModel);
+        std::unique_ptr<Anopheles::EmergenceRateEstimator> estimator;
+        if (util::ModelOptions::option(util::USE_EXACT_NV0_SOLVER))
+            estimator.reset(new Anopheles::EmergenceRateSolver(*anophModel, populationSize));
+        else
+            estimator.reset(new Anopheles::EmergenceRateAdaptiveFitter(*anophModel));
```
b-s-code: std::make_unique might be preferable to `new` here.

model/util/ModelOptions.cpp
```diff
@@ -118,8 +119,9 @@ namespace OM { namespace util {
 
     std::bitset<NUM_OPTIONS> ModelOptions::getBaseModelOptions()
     {
-        // The "base" model is very simple: *no* model options are turned on.
+        // The "base" model is very simple, but still uses current defaults.
         bitset<NUM_OPTIONS> allOptionsOff;
+        allOptionsOff.set (USE_EXACT_NV0_SOLVER);
```
b-s-code: I think there's an opportunity to do something a little more expressive here.

One might expect that a bitset called `allOptionsOff` has all zeroes.

test/scenarioExactNv0.xml
```diff
@@ -0,0 +1,319 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<!-- This file is part of OpenMalaria.
+Copyright © 2005-2025 Swiss Tropical and Public Health Institute
+Copyright © 2005-2015 Liverpool School Of Tropical Medicine
+Copyright © 2020-2025 University of Basel
+Copyright © 2025 The Kids Research Institute Australia
```
b-s-code: 2026

test/scenarioModelNameNoOverrides.xml
```diff
@@ -271,7 +271,9 @@
     <!-- Here you can specify additional model parameters, or model variants. -->
     <model>
         <ModelName name="base"/>
-
+        <ModelOptions>
+          <option name="USE_EXACT_NV0_SOLVER" value="false"/>
+        </ModelOptions>
```
b-s-code: One intended (but maybe not documented) purpose of this test is to ensure OpenMalaria doesn't crash in the case where no model options are overridden and no parameters are overridden.

We lose that coverage by adding this code.

Perhaps we could modify the expected output txt for this test instead of modifying to the scenario?

model/util/ModelOptions.cpp
```diff
@@ -136,6 +138,9 @@ namespace OM { namespace util {
         defaultOptSet.set (INDIRECT_MORTALITY_FIX);
         defaultOptSet.set (HEALTH_SYSTEM_MEMORY_FIX);
 
+        // Current default. Disable to use the legacy adaptive fitter.
+        defaultOptSet.set (USE_EXACT_NV0_SOLVER);
```
b-s-code: Since this method returns legacy model options, shouldn't USE_EXACT_NV0_SOLVER be off?

b-s-code (commented):
Overall approach looks great to me and it seems clear to me that this PR is a good change.

I have given many minor remarks.

My most important point of feedback is on ModelOptions::getBaseModelOptions(), and it's more a readability concern than anything else.

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: done

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: Using constexpr instead

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: I have removed the linear + smoothness penalty path completely. This code is gone.

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: This code is gone, too.

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: Gone as well.

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: It was from a different branch, I have removed this comment.

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: This is standard practice for the GSL solver. This allows the user to pass any type of parameters to the function.

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: done !

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: updated !

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: Removed

model/Transmission/Anopheles/CalcNvO.cpp
acavelan: Added an exception

model/Transmission/Anopheles/EmergenceRateEstimator.h
acavelan: done

model/Transmission/Anopheles/EmergenceRateEstimator.h
acavelan: Only the legacy solver needs to iterate. The new solver calculates the unique solution, so it must work on the first pass and therefore it always returns false (no need to iterate).

model/Transmission/transmission.h
acavelan: done

model/util/ModelOptions.cpp
acavelan: I have renamed the model option to USE_LEGACY_NV0_SOLVER so we can keep it allOptionsOff while having the new solver as the default.

test/scenarioModelNameNoOverrides.xml
acavelan: I have renamed the model option to "USE_LEGACY_NV0_SOLVER" so this scenario does not need any override and now uses the new solver. I have updated the expected outcome for this scenario (and only for this one).

model/util/ModelOptions.cpp
acavelan: Removed, see above!

acavelan:
The main changes are:
- Removed the complicated linear solver with smoothness and ridge penalty since it is no longer needed
- Renamed the model option to USE_LEGACY_NV0_SOLVER so that it is false by default and included in the base model

model/Transmission/Anopheles/EmergenceRateEstimator.h
b-s-code: Makes sense, thanks for adding comment to that effect.

model/util/ModelOptions.cpp
b-s-code: Comment now outdated, but I'm satisfied new arrangement is good :+1:

test/scenarioModelNameNoOverrides.xml
b-s-code: Comment now outdated, but I'm satisfied new arrangement is good :+1:

b-s-code (approved):
Looks good!

I suppose it might be best not to modify https://github.com/OpenMalaria-Org/openmalaria/wiki/ScenarioTransmission#limits-of-the-model now, since the default method has been changed back to the adaptive fitter.

PR #452, opened 2026-02-17
feat: Build with C++20.

b-s-code:
# Problem

Some previous PRs prepared to switch to C++20.  https://github.com/OpenMalaria-Org/openmalaria/pull/431 https://github.com/OpenMalaria-Org/openmalaria/pull/426

Now that switching to C++20 is unblocked, we wish to proceed such that we can get access to latest language features.

This is not absolutely critical/blocking any other important tasks but rather represents a reasonable improvement.

## Solution

Tell CMake to build with C++20.

### Additional cleanup

While reviewing the top level CMakeLists.txt file, I also noticed some other settings which appeared to be a workaround to the fact that XSD headers used auto_ptr, which was removed in C++17.  These no longer appear to be necessary and so I have removed them, presumably XSD headers have been updated.

## Testing

Tested locally (Debian, GCC, X86_64): built ok, tests passed.
Tested on Setonix with GCC: built ok, tests passed.
CI builds: built ok, tests passed.

## Documentation

Only hits in the wiki for "C++17" or "CPP17: are in changelog, so no need for any wiki updates.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ set (CMAKE_MODULE_PATH ${CMAKE_SOURCE_DIR}/cmake)
 
 # -----  Compile options  -----
 
-set(CMAKE_CXX_STANDARD 17)
+set(CMAKE_CXX_STANDARD 20)
 
 set (OM_STD_LIBS)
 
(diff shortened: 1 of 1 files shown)


b-s-code:
@acavelan Can I please get a review?  I'm happy with testing done from my side.  If there were any other platforms you think should be covered lmk.

@melissapenny @nakul7 FYI

acavelan (approved):
All good, thanks for checking XSD compatibility. I added this a while ago and I am happy to get rid of it since it is no longer required.

PR #454, opened 2026-03-19
feat: Add error output to suggest possible causes when scenario parsing fails

b-s-code:
# Problem

A misplaced/misnamed/unlocatable XSD file is not easily identified by new users, from existing OpenMalaria stderr output.

This is a mostly a quality of life issue, where very experienced OpenMalaria users can generally infer the nature of the issue from existing stderr output contents.  But only because they've already spent the time to learn what the symptoms mean.

## Solution

Much of the code that gets called when parsing the scenario is **generated** (by xsdcxx) C++ code.

To avoid getting too involved with such generated code, I looked at what could be done in relevant code **within** OpenMalaria's source tree.

The solution presented in this PR is to simply print additional output to stderr when parsing fails, suggesting 2 possible causes to the user.

Additionally, in this solution we distinguish between failure modes in order to provide more targeted feedback to the user.

## Testing

I tested before/after for 3 cases:

1. Scenario conforms to schema, schema is appropriately named/located.
2. Scenario conforms to schema, schema is wrongly named/located.
3. Scenario does not conform to schema, schema is appropriately named/located.

### Scenario conforms to schema, schema is appropriately named/located.

This is the happy path.  No change in output occurs.

### Scenario conforms to schema, schema is wrongly named/located.

Before:

```
:0:0 warning: unable to open primary document entity 'scenario_48.xsd'
:22:234 error: no declaration found for element 'om:scenario'
:22:234 error: attribute '{http://www.w3.org/2000/xmlns/}om' is not declared for element 'om:scenario'
:22:234 error: attribute '{http://www.w3.org/2000/xmlns/}xsi' is not declared for element 'om:scenario'
:22:234 error: attribute 'name' is not declared for element 'om:scenario'
:22:234 error: attribute 'schemaVersion' is not declared for element 'om:scenario'
:22:234 error: attribute '{http://www.w3.org/2001/XMLSchema-instance}schemaLocation' is not declared for element 'om:scenario'
:29:66 error: no declaration found for element 'demography'
:29:66 error: attribute 'maximumAgeYrs' is not declared for element 'demography'

[...]

:341:55 error: attribute 'value' is not declared for element 'group'
:345:43 error: no declaration found for element 'computationParameters'
:345:43 error: attribute 'iseed' is not declared for element 'computationParameters'
OpenMalaria: No such file or directory
```

After:

```
Parsing scenario file failed.
        Likely the XSD schema file is not present at an expected location/with expected filename.
OpenMalaria: No such file or directory
```

### Scenario does not conform to schema, schema is appropriately named/located.

(For this test case, I used 48.0 schema, instead of scenario_current.xsd.)

Before:

```
:192:46 error: missing required attribute 'propInfected'
:192:46 error: missing required attribute 'propInfectious'
:222:23 error: missing required attribute 'minInfectedThreshold'
:243:44 error: missing required attribute 'propInfected'
:243:44 error: missing required attribute 'propInfectious'
:250:23 error: missing required attribute 'minInfectedThreshold'
:271:46 error: missing required attribute 'propInfected'
:271:46 error: missing required attribute 'propInfectious'
:278:23 error: missing required attribute 'minInfectedThreshold'
```

After:

```
:192:46 error: missing required attribute 'propInfected'
:192:46 error: missing required attribute 'propInfectious'
:222:23 error: missing required attribute 'minInfectedThreshold'
:243:44 error: missing required attribute 'propInfected'
:243:44 error: missing required attribute 'propInfectious'
:250:23 error: missing required attribute 'minInfectedThreshold'
:271:46 error: missing required attribute 'propInfected'
:271:46 error: missing required attribute 'propInfectious'
:278:23 error: missing required attribute 'minInfectedThreshold'
Parsing scenario file failed.
        Likely the XSD schema is named/located correctly but the scenario does not conform to the schema.
```

## Documentation

No wiki updates or other documentation changes needed for this PR.

### DIFF
--- a/model/main.cpp
+++ b/model/main.cpp
@@ -310,15 +310,20 @@ int main(int argc, char* argv[])
             exitStatus = e.getCode();
         }
     } catch (const ::xsd::cxx::tree::exception<char>& e) {
-        // e.what() returns a concise error message e.g. "instance document parsing failed".
-        // Whereas inserting e into a stream inserts the (possibly long and repetitive) list of error messages arising from the parsing attempt.
-        std::cerr << "XSD error: " << e.what() << '\n' << e << endl;
-
-        // We print this *after* the above since users are more likely to take note of the tail of stderr/stdout output than the head.
         if (errno == 2) // Errno value 2 corresponds to "No such file or directory" i.e. this is not a mismatch between scenario contents and schema rules.
+        {
+            // We don't print the content of the xsdcxx exception since the error messages are not relevant to the user in this case.
             std::cerr << "Parsing scenario file failed.\n\tLikely the XSD schema file is not present at an expected location/with expected filename." << std::endl;
+        }
         else
+        {
+            // e.what() returns a concise error message e.g. "instance document parsing failed".
+            // Whereas inserting e into a stream inserts the (possibly long and repetitive) list of error messages arising from the parsing attempt.
+            std::cerr << "XSD error: " << e.what() << '\n' << e << endl;
+
+            // We print this *after* the above since users are more likely to take note of the tail of stderr/stdout output than the head.
             std::cerr << "Parsing scenario file failed.\n\tLikely the XSD schema is named/located correctly but the scenario does not conform to the schema." << std::endl;
+        }
         exitStatus = OM::util::Error::XSD;
     } catch (const OM::util::checkpoint_error& e) {
         std::cerr << "Checkpoint error: " << e.what() << endl;
--- a/model/main.cpp
+++ b/model/main.cpp
@@ -304,35 +304,41 @@ int main(int argc, char* argv[])
     } catch (const OM::util::cmd_exception& e) {
         if( e.getCode() == 0 ){
             // this is not an error, but exiting due to command line
-            cerr << e.what() << "; exiting..." << endl;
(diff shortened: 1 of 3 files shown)


b-s-code:
@acavelan may I please get a review?

@melissapenny @nakul7 FYI

acavelan:
This is a really good idea.

It would be even better if we could detect that the file does not exist before trying to parse the document. I believe users get confused because they get 300 lines of errors when this happens. They will probably miss the new helper text at the top.

If not possible, perhaps the new error description should appear at the end?

b-s-code:
@acavelan can I please get another review?

> It would be even better if we could detect that the file does not exist before trying to parse the document. I believe users get confused because they get 300 lines of errors when this happens. They will probably miss the new helper text at the top.

>If not possible, perhaps the new error description should appear at the end?

I moved the most important info to the bottom of the error output, I think that's a great suggestion.

I've now managed to distinguish between the schema file not being found vs. the scenario not conforming to the schema.  The PR description is updated with new output examples.

I wasn't able to detect that the file does not exist *before* parsing is attempted, but I managed to hide all the error spam from the user when that failure mode occurs.

acavelan (approved):
All good, thanks!

PR #456, opened 2026-06-09
Schema 49

acavelan:
Version number increased to 49

- Removed unnecessary outputs
- Added `<option name="USE_EXACT_NV0_SOLVER" value="true"/>` to example_scenario.xml to enable the new emergence rate calculation for new users while keeping the old method as the default
- Edited: https://github.com/OpenMalaria-Org/openmalaria/wiki/ScenarioTransmission
- Updated Docker files


b-s-code (commented):
Looks good, only suggestion is adding a Dockerfile for v49.

docker/49.0
```diff
@@ -0,0 +1,30 @@
...
+WORKDIR /
+
+ARG OM_BRANCH=main
+ARG OM_VERSION=49
+
+# schema-48.0 release does not have changes required to build OM on linux/arm64.
+RUN apt-get update && apt-get install -y --no-install-recommends \
+  build-essential ca-certificates git libz-dev python3 cmake libgsl-dev libxerces-c-dev xsdcxx libboost-dev \
+  && rm -rf /var/lib/apt/lists/* \
+  && git clone --depth 1 --branch ${OM_BRANCH} https://github.com/SwissTPH/openmalaria.git \
```
b-s-code: I imagine this works fine but would be worth updating SwissTPH to OpenMalaria-Org.

util/generateDoc.py
```diff
@@ -683,8 +683,8 @@ def main():
             w.p('Comments are welcome but be warned that edits will most likely be lost.')
             w.pn('')
             w.p('The script used to do this is')
-            w.p('[generateDoc.py](https://github.com/SwissTPH/openmalaria.tools/blob/master/openmalaria/tools/generateDoc.py),')
-            w.p(', found in the [openmalaria.tools repository](https://github.com/SwissTPH/openmalaria.tools).')
+            w.p('[generateDoc.py](https://github.com/OpenMalaria-Org/openmalaria.tools/blob/master/openmalaria/tools/generateDoc.py),')
+            w.p(', found in the [openmalaria.tools repository](https://github.com/OpenMalaria-Org/openmalaria.tools).')
```
b-s-code: We still want SwissTPH for these URLs.

b-s-code (commented):
The decision to update many URLs is a good one. I found a couple links that broke with the find and replace (openmalaria.tools repo) but otherwise looks good.

b-s-code (approved):
Looks ready to me.

PR #458, opened 2026-08-04
feat: Add a named model for Molineaux model.

b-s-code:
# Problem 

When a user wants to use the Molineaux model, they must somehow obtain a parameterisation of it and a list of appropriate model options.

The problem this PR solves is eliminating the need to track down such model options and parameter values in order to use the Molineaux model.

It's hoped this will improve reproducibility and save on communication, by making clear the provenance of the values used.

## Solution

In https://github.com/OpenMalaria-Org/openmalaria/pull/419, the ability to hardcode a collection of model options and parameter values, identifiable by some string name was added.

Thus users can use a pre-defined collection of model options and parameter values, if they reference that collection by name in their scenario. This PR adds such a collection of values under the name `molineaux_original`.

A user can select this pre-defined collection of model options/parameters by specifying the value `molineaux_original` for the `name` attribute on the `<ModelName>` tag:

```xml
    <model>
        <ModelName name="molineaux_original"/>
        <!-- <ModelOptions> node not required unless overriding some value. -->
        ...
        <!-- <parameters> node not required unless overriding some value. -->
   </model>
```

The collection of model options and values being added is consistent with those values found in the papers/accompanying workflow code referenced below.

Points of note:

- Some model options are not hardcoded since doing so would now be redundant (see ModelOptions.cpp comments for details).
- There are some values that need to be specified in the scenario XML in order to use the Molineaux model, which are not packaged up with the "molineaux_original" in this PR. In particular `<weight>` under `<human>` under `<model>`. TODO : is that all?
- TODO : consider HEALTH_SYSTEM_MEMORY_FIX... would have been on (implicitly) in original scenarios. But it's not here...

## Testing

Added one new test to cover the newly added combination of ModelName/Parameters/ModelOptions.

All existing tests continue to pass.

#### Wiki

- [ ] Update https://github.com/OpenMalaria-Org/openmalaria/wiki/ModelWithinHost#model-options-and-parameters to reflect changes in this PR (TODO).

#### References

| Paper | Year | DOI | Zenodo code URL | XML sample on GitHub |
|---|---|---|---|---|
| The influence of biological, epidemiological, and treatment factors on the establishment and spread of drug-resistant Plasmodium falciparum | 2022 | https://doi.org/10.7554/eLife.77634 | https://zenodo.org/records/6046967 | https://github.com/ThieryM95/Drug_resistance_workflow/blob/45ddd6a3e2523009387e7e4a8fab5c2221bf8d5c/wf_esthablishment/SIM_FOLDER/Esthablishment_long.xml |
| Seasonal malaria chemoprevention and the spread of Plasmodium falciparum quintuple-mutant parasites resistant to sulfadoxine-pyrimethamine: a modelling study | 2024 | https://doi.org/10.1016/s2666-5247(24)00115-0 | https://zenodo.org/records/7244732 | https://github.com/ThieryM95/SP_resistance_workflow/blob/f976158a060b535bbbc1416394b432b16c4efd56/SIM_FOLDER/Propylaxis_new_dosage_AQ_DEAQ.xml |


b-s-code:
@acavelan @melissapenny @nakul7 This PR adds one parameterisation of the Molineaux model to OpenMalaria as a pre-set named model, so that users don't have to explicitly list model options and parameter values.

Please share any initial feedback on this change. I'll also see if I can get some user feedback.

model/util/ModelNameProvider.h
```diff
@@ -36,7 +36,8 @@ namespace OM
         enum class ModelNames
         {
             none, // Represents the case where input XML contains no model name.
-            base
+            base,
+            molineaux_original
```
acavelan, line 40: Is there a reason to use "molineaux_original" instead of "molineaux"?

model/util/ModelOptions.cpp
acavelan, line 1: HEALTH_SYSTEM_MEMORY_FIX should be off as it was introduced very recently. It should be reset alongside INDIRECT_MORTALITY_FIX.

PR #459, opened 2026-08-04
Avoid checkpointing null decay samples

acavelan:


b-s-code (approved):
Looks like a good solution.

